PR #11, opened 2015-02-12
Make key wrap more generic

csstaub:
This makes the key wrap implementation the `josecipher` subpackage more generic, so that it can be used with block ciphers other than AES.


csstaub:
R: @dgalling @alokmenghrajani

alokmenghrajani:
lgtm

PR #16, opened 2015-02-27
shared.go: "RCDSA" -> "ECDSA"

dlg-yahoo:
Patch is released into the public domain.

(Yahoo's lawyers will need to talk to your lawyers about a CLA, but I'm authorized to release utterly trivial patches into the public domain.)


alokmenghrajani:
Sorry, I missed this PR. Unsure about taking pull requests without a CLA, someone else put up a "similar" fix, I'll merge that one to save time.

Issue #17, opened 2015-03-16
Signature verification chicken-and-egg problem


rapropos:
I have a situation where information included in an incoming payload is needed to know which key to use to verify the signature on a request, and cannot figure out how to achieve this with the current API. Would there be a downside to exposing the JsonWebSignature's payload field in some fashion that does not yet require knowing the key needed for verification? I am looking to do something like:
```
json.Unmarshal(jwo.Payload(), &req)
database.FindId(req.SubmitterId, &submitter)
jwo.Verify(submitter.SigningKey)
```
Thank you for your consideration.

csstaub:
It would be preferable to store this information inside a header value, such as the `kid` (key id) header field. I'm currently working on exposing these header values in my `cs/jwk` branch. Selecting the verification key based on the payload seems like an anti-pattern to me, I'd rather not expose the payload in a way that would allow callers to skip verification. Is there any reason you are unable to use header values for key selection?

rapropos:
I was not aware that these header fields could be co-opted to contain arbitrary business logic data. The description of the `kid` field in the draft spec seemed to suggest that it was intended for a situation where a single signer has multiple keys. In my application, a server needs to respond to many different signers, and the verification process consists of making sure that each successive request is signed with the key that was registered for that signer during an initial registration process.

I will try using a header field and your experimental branch.

rapropos:
So far I have been unsuccessful attempting to set the KeyId field in such a fashion so that it arrives at the server. I marshal my payload object into a `[]byte`, pass that to `Sign()`, set `Signatures[0].Header.KeyId` on the resulting signed object, call `FullSerialize` and send the result over the wire. When it arrives on the server side, the `KeyId` field of the header is blank. A cursory inspection of the `JsonWebSignature` object before it is serialized shows both a `KeyId` field in the `Signatures` array (which contains the proper value I set) and a `KeyId` field inside the `rawHeader` structure (which is blank). I see that the comments around the definition of the `JoseHeader` type call it "read-only", which may be the crux of my problem. Is there a supported way to get that 'kid' header field set via the go-jose library in addition to reading it?

csstaub:
[...]

PR #19, opened 2015-03-18
first pass at JWT private keys and KeyId for signing

rapropos:
Here is an attempt at adding RSA and ECDSA private key support to the JWK code, along with trying to get KeyId specified in the `kid` field of a JWS signing key passed through the signature headers so that it is available to recipients between the parsing and verification stage. Intended to address #17. I hope this may be of some use to folks, and apologies for any ugliness therein.

### DIFF
--- a/jwk.go
+++ b/jwk.go
@@ -23,23 +23,30 @@ import (
 	"encoding/binary"
 	"encoding/json"
 	"fmt"
+	"math/big"
 	"reflect"
 )
 
 // rawJsonWebKey represents a public or private key in JWK format, used for parsing/serializing.
 type rawJsonWebKey struct {
-	// TODO(cs): Add support for private keys.
 	Kty string      `json:"kty,omitempty"`
+	Kid string      `json:"kid,omitempty"`
 	Crv string      `json:"crv,omitempty"`
 	X   *byteBuffer `json:"x,omitempty"`
 	Y   *byteBuffer `json:"y,omitempty"`
 	N   *byteBuffer `json:"n,omitempty"`
 	E   *byteBuffer `json:"e,omitempty"`
+	// private key fields, EC uses D only
+	// if D is nil, key is public only
+	D *byteBuffer `json:"d,omitempty"`
+	P *byteBuffer `json:"p,omitempty"`
+	Q *byteBuffer `json:"q.omitempty"`
 }
 
 // JsonWebKey represents a public or private key in JWK format.
 type JsonWebKey struct {
 	key interface{}
+	kid string
 }
 
 func (k *JsonWebKey) MarshalJSON() ([]byte, error) {
@@ -49,9 +56,15 @@ func (k *JsonWebKey) MarshalJSON() ([]byte, error) {
 		raw.fromEcPublicKey(key)
 	case *rsa.PublicKey:
 		raw.fromRsaPublicKey(key)
+	case *ecdsa.PrivateKey:
+		raw.fromEcPrivateKey(key)
+	case *rsa.PrivateKey:
+		raw.fromRsaPrivateKey(key)
 	default:
 		return nil, fmt.Errorf("square/go-jose: unkown key type '%s'", reflect.TypeOf(key))
 	}
+
+	raw.Kid = k.kid
 
 	return json.Marshal(raw)
 }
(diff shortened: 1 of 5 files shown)


csstaub:
Hi, this looks great! Did you take a look at the [`CONTRIBUTING.md`](https://github.com/square/go-jose/blob/master/CONTRIBUTING.md) file in the repo, and in particular, have you signed our [Individual Contributor License Agreement](https://docs.google.com/a/squareup.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?ndplr=1&formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ) linked in there?

jwk.go
```diff
@@ -145,3 +166,56 @@ func (key *rawJsonWebKey) fromEcPublicKey(pub *ecdsa.PublicKey) error {
...
+	rv := new(rsa.PrivateKey)
+	pubkey, err := jwk.rsaPublicKey()
+	if err != nil {
+		return nil, err
+	}
+	rv.PublicKey = *pubkey
+	rv.D = jwk.D.bigInt()
+	rv.Primes = make([]*big.Int, 2)
+	rv.Primes[0] = jwk.P.bigInt()
+	rv.Primes[1] = jwk.Q.bigInt()
```
csstaub: Would be preferable to build a struct here directly, and avoid calling `jwk.rsaPublicKey()`, e.g like:

```Go
priv := &rsa.PrivateKey{
  PublicKey: rsa.PublicKey{
    N: jwk.N.bigInt(),
    E: jwk.E.toInt(),
  },
  D: jwk.D.bigInt(),
  Primes: []*big.Int{
    jwk.P.bigInt(),
    jwk.Q.bigInt(),
  },
}
```

Also, this should set the `Precomputed` field (type `rsa.PrecomputedValues`) if DP / DQ / QI values are present.

jwk.go
```diff
@@ -145,3 +166,56 @@ func (key *rawJsonWebKey) fromEcPublicKey(pub *ecdsa.PublicKey) error {
...
+	err = rv.Validate()
+	return rv, err
+}
+
+func (jwkp *rawJsonWebKey) fromRsaPrivateKey(rsa *rsa.PrivateKey) {
+	var jwk rawJsonWebKey
+	jwk.fromRsaPublicKey(&rsa.PublicKey)
+
+	jwk.D = newBuffer(rsa.D.Bytes())
+	// TODO: do we need to ensure len(Primes) >= 2 and/or deal with more than 2?
```
csstaub: As far as I can see, the JWK spec does not mention how to represent RSA keys with more than two primes. The library should probably check if a key has more than two primes and reject it.

jwk.go
```diff
@@ -145,3 +166,56 @@ func (key *rawJsonWebKey) fromEcPublicKey(pub *ecdsa.PublicKey) error {
...
+		return nil, fmt.Errorf("square/go-jose: invalid ECDSA private key, missing D")
+	}
+
+	rv := new(ecdsa.PrivateKey)
+	pubkey, err := jwk.ecPublicKey()
+	if err != nil {
+		return nil, err
+	}
+	rv.PublicKey = *pubkey
+	rv.D = jwk.D.bigInt()
```
csstaub: Same thing here as for RSA private keys above.

jwk.go
```diff
@@ -23,23 +23,30 @@ import (
...
-	// TODO(cs): Add support for private keys.
 	Kty string      `json:"kty,omitempty"`
+	Kid string      `json:"kid,omitempty"`
 	Crv string      `json:"crv,omitempty"`
 	X   *byteBuffer `json:"x,omitempty"`
 	Y   *byteBuffer `json:"y,omitempty"`
 	N   *byteBuffer `json:"n,omitempty"`
 	E   *byteBuffer `json:"e,omitempty"`
+	// private key fields, EC uses D only
+	// if D is nil, key is public only
```
csstaub: Comment is somewhat confusing, needs better explanation.

rapropos:
Contractor license agreement is signed. Regarding filling in the entire struct directly versus calling the various public key functions to do their part, I thought it was more DRY to do it this way, but I can see your position as well, especially since the public key part isn't going to change in the future, so I will change things to reflect your comments. Thank you for taking the time to review the code.

csstaub:
Great, thanks! Merging this into my experimental branch. Want to work on this some more before merging to master though. Also, I think it might make sense to make the fields on `JsonWebKey` public so that callers can construct their own keys.

rapropos:
Thank you for your consideration, and please don't hesitate to let me know if I can be of further assistance. Not sure on the etiquette, but have provisionally closed #17 as this functionality solves my use case.

Issue #7, opened 2014-12-23
Expose header values on objects


csstaub:
Expose header values to e.g. retrieve a JWK embedded in a JWS header.

csstaub:
Merged #20, exposes some basic header values now.

Issue #3, opened 2014-12-19
Support JWK


bifurcation:
It would be helpful for this library to have support for processing JWK-formatted keys independent of JWS / JWE.  For example:
https://github.com/bifurcation/gose/blob/master/jwk.go

bifurcation:
I updated the GOSE repo to mark it as CC0.

csstaub:
Merged #20, which now has basic support for JWK.

Issue #2, opened 2014-12-19
Improve JWS key management for verification


bifurcation:
The current "supply a key to verify with" model for verification is almost always a bad fit.

If a JWS object contains a "jwk" header, then it already has enough information for a verifier to be able to check the signature.  In this case, it should not be necessary to supply a public key to a verifier constructor.

If a JWS object does not contain a "jwk" header, then the verifier should be identifying the key for verification based on other header fields, e.g., "kid".  In this case, a set of keys with attributes would be an appropriate input to a verifier factory, instead of an individual key.

It might be useful simply to add a Verify() and/or Verify([]KeyType) methods to JwsObject.  For example:
https://github.com/bifurcation/gose/blob/master/jws.go#L211

csstaub:
Hi @bifurcation, I'm working on this (and JWK support) in a separate branch right now: https://github.com/square/go-jose/tree/cs/jwk. Feedback is welcome :).

bifurcation:
Hey @csstaub, thanks for the ping, I'll take a look.  BTW, once you've got JWK and JWS working, if you want to take a stab at integrating it with [boulder](https://github.com/letsencrypt/boulder), that would be awesome.

csstaub:
Merged #20, headers now expose JWK and can be used for verification.

PR #20, opened 2015-03-19
Basic support for JWK and headers

csstaub:
This adds basic support for the JWK format, and allows for reading (some, not all) header values from parsed messages. Please take a look.

R: @dgalling @sul3n3t
CC: @rapropos @bifurcation

### DIFF
--- a/asymmetric.go
+++ b/asymmetric.go
@@ -85,6 +85,9 @@ func newRSASigner(sigAlg SignatureAlgorithm, privateKey *rsa.PrivateKey) (recipi
 
 	return recipientSigInfo{
 		sigAlg: sigAlg,
+		publicKey: &JsonWebKey{
+			key: &privateKey.PublicKey,
+		},
 		signer: &rsaDecrypterSigner{
 			privateKey: privateKey,
 		},
@@ -119,6 +122,9 @@ func newECDSASigner(sigAlg SignatureAlgorithm, privateKey *ecdsa.PrivateKey) (re
 
 	return recipientSigInfo{
 		sigAlg: sigAlg,
+		publicKey: &JsonWebKey{
+			key: &privateKey.PublicKey,
+		},
 		signer: &ecDecrypterSigner{
 			privateKey: privateKey,
 		},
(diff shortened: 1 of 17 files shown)


csstaub:
(This should address issues #2, #3 and #7)

asymmetric.go
```diff
@@ -219,7 +226,7 @@ func (ctx rsaDecrypterSigner) signPayload(payload []byte, alg SignatureAlgorithm
 	case RS512, PS512:
 		hash = crypto.SHA512
 	default:
-		return signatureInfo{}, ErrUnsupportedAlgorithm
+		return Signature{}, ErrUnsupportedAlgorithm
```
dgalling: Should this return `nil, ErrUnsupportedAlgorithm`?

csstaub: Can't return `nil` here, return value is not a pointer.

dgalling: Ah, duh.

jwe.go
```diff
@@ -116,12 +117,19 @@ func parseEncryptedFull(input string) (*JsonWebEncryption, error) {
 		return nil, err
 	}
 
-	obj := &JsonWebEncryption{}
-	obj.original = &parsed
-	obj.unprotected = parsed.Unprotected
+	return parsed.sanitized()
+}
+
+func (parsed *rawJsonWebEncryption) sanitized() (*JsonWebEncryption, error) {
```
sul3n3t: It is private, but I think a method-level comment might be helpful.

jwk_test.go
```diff
@@ -17,40 +17,240 @@
...
+		},
+		nil,
+	}
+
+	for _, key := range keys {
+		jwk := JsonWebKey{Key: key}
+		_, err := jwk.MarshalJSON()
 		if err == nil {
-			t.Error("parser incorrectly parsed invalid key", jwk)
+			t.Error("managed to serialize invalid key")
```
sul3n3t: For all the looped test cases, it would be helpful to output which case failed with the error message.

sul3n3t:
There's a lot here. I don't have too many comments. Aside from those I already added, consider changing the signatures for methods like `fromRsaPublicKey` to look like constructors and not method receivers. Comments for the sanitized methods would be helpful.

Looks pretty good. :+1:

PR #23, opened 2015-05-05
Add dump command to jose-util

csstaub:
This adds a `dump` command to `jose-util`, which parses a message and then dumps the pretty-printed full serialization format. Meant to be used for debugging. Can also update this in the future to print a nicer, more human-readable dump of the input message.

R: @dgalling

### DIFF
--- a/jose-util/main.go
+++ b/jose-util/main.go
@@ -17,18 +17,20 @@
 package main
 
 import (
+	"encoding/json"
 	"fmt"
-	"github.com/codegangsta/cli"
-	"github.com/square/go-jose"
 	"io/ioutil"
 	"os"
+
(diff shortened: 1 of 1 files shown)


dgalling:
Is there no easier way to do this than parsing, serializing, unmarshaling, and then marshaling?

csstaub:
Not without changing some internals in the `jose` package, e.g. giving it an option to pretty-print when producing the full serialization format. That would avoid the re-marshaling step at the end. Initial parsing/serializing is required to make sure we can support both inputs formats.

dgalling:
Ok, let's do it this way for now, but file an issue to support pretty-printing in the full serialization.

Issue #22, opened 2015-05-03
Sign with no Protected header in Flattened JWS JSON Serialization syntax


jsha:
By my reading of the code, it looks like the unprotected header field is not supported in [Flattened JWS JSON Serialization](https://tools.ietf.org/html/draft-ietf-jose-json-web-signature-41#page-22). I'd like to add support: We are using it in https://github.com/letsencrypt/boulder to indicate which key was used to sign a message, and we'd like to port Boulder to use go-jose. Any objections?

Thanks,
Jacob

csstaub:
Hi @jsha, no objections to adding support! Pull requests are definitely welcome :)

jsha:
After further investigation, the problem was not the unprotected header but the lack of any Protected header. Fix in https://github.com/square/go-jose/pull/27.

PR #27, opened 2015-05-08
Fix sign/verify when protected header is absent.

jsha:
Previously, if there was no protected header, we would attempt to serialize a
nil pointer. json.Marshal would happily serialize it as the literal string
"null", which would cause an incorrect signature or a verify error.

Instead, if there is no protected header, explicitly set the protected part of
the input to be signed to the empty string.

Fixes https://github.com/square/go-jose/issues/22

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -48,6 +48,12 @@ func mustSerializeJSON(value interface{}) []byte {
 	if err != nil {
 		panic(err)
 	}
+	// Note: It's not possible to directly check whether the data pointed at by an
+	// interface is a nil pointer, so we do this hacky workaround.
+	// https://groups.google.com/forum/#!topic/golang-nuts/wnH302gBa4I
+	if string(out) == "null" {
+		panic("Tried to serialize a nil pointer.")
+	}
 	return out
 }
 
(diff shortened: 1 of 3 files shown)


csstaub:
Mandatory question: Have you signed our [Individual Contributor License Agreement](https://docs.google.com/a/squareup.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?ndplr=1&formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ)?

jsha:
I have now!

encoding.go
```diff
@@ -48,6 +48,12 @@ func mustSerializeJSON(value interface{}) []byte {
 	if err != nil {
 		panic(err)
 	}
+	// Note: It's not possible to directly check whether the data pointed at by an
+	// interface is a nil pointer, so we do this hacky workaround.
+	// https://groups.google.com/forum/#!topic/golang-nuts/wnH302gBa4I
+	if string(out) == "null" {
```
csstaub: I think the comment here should explain why we're doing this (to prevent accidental serialization of a value). Without the context of the pull request it's not immediately obvious why this is here.

jsha: Good point, expanded the comment.

csstaub:
Looks good. Thanks for fixing this!

PR #30, opened 2015-05-08
Check key size on ECDSA signing

csstaub:
Check key size on ECDSA signing, fixes a panic when using a key with incorrect algorithm (e.g. using ES256 signature with P-384 key).

R: @dgalling

### DIFF
--- a/asymmetric.go
+++ b/asymmetric.go
@@ -404,19 +404,24 @@ func (ctx ecDecrypterSigner) decryptKey(headers rawHeader, recipient *recipientI
 
 // Sign the given payload
 func (ctx ecDecrypterSigner) signPayload(payload []byte, alg SignatureAlgorithm) (Signature, error) {
-	var keySize int
+	var expectedBitSize int
 	var hash crypto.Hash
 
 	switch alg {
 	case ES256:
-		keySize = 32
+		expectedBitSize = 256
 		hash = crypto.SHA256
 	case ES384:
-		keySize = 48
+		expectedBitSize = 384
 		hash = crypto.SHA384
 	case ES512:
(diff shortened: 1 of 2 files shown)


asymmetric.go
```diff
@@ -430,13 +435,18 @@ func (ctx ecDecrypterSigner) signPayload(payload []byte, alg SignatureAlgorithm)
 		return Signature{}, err
 	}
 
+	keyBytes := curveBits / 8
+	if curveBits%8 > 0 {
```
dgalling: nit: spaces around operators

csstaub: Nope, `gofmt` formats it like this.

asymmetric.go
```diff
@@ -430,13 +435,18 @@ func (ctx ecDecrypterSigner) signPayload(payload []byte, alg SignatureAlgorithm)
...
+		keyBytes += 1
+	}
+
 	rBytes := r.Bytes()
-	rBytesPadded := make([]byte, keySize)
-	copy(rBytesPadded[keySize-len(rBytes):], rBytes)
+	rBytesPadded := make([]byte, keyBytes)
+	copy(rBytesPadded[keyBytes-len(rBytes):], rBytes)
```
dgalling: A comment explaining how this padding works would be nice (that `r.Bytes()` is big-endian and needs zeroes prepended to make it `keyBytes` long).

asymmetric.go
dgalling: That's unfortunate.

asymmetric.go
csstaub: done

PR #32, opened 2015-05-09
Fix Verify for Flattened Protected Private Headers

jsha:
In Flattened representation, if the Protected header contained private field
names, the signature wouldn't verify. This change retains the original data for
the Protected header field to enable proper verification.


csstaub:
Looks good!

PR #33, opened 2015-05-11
Includes this repo in our open source bug bounty program.

alokmenghrajani:


csstaub:
Hah, just added that too.

alokmenghrajani:
We should agree on bug-bounty.md vs BUG-BOUNTY.md and stick to the same naming convention?

csstaub:
I prefer upper-case since all the other text files are upper-case too (LICENSE, README.md, CONTRIBUTING.md), but I don't think it really matters

PR #37, opened 2015-06-29
Add support for JWK thumbprints

bifurcation:
The IETF draft defining [JWK thumbprints](http://tools.ietf.org/html/draft-ietf-jose-jwk-thumbprint) is almost an RFC.  It would be helpful for this library to support thumbprints in the standard format.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -131,6 +131,22 @@ func newBuffer(data []byte) *byteBuffer {
 	}
 }
 
+func newZeroPaddedBuffer(data []byte, length int) *byteBuffer {
+	if len(data) >= length {
+		return newBuffer(data)
+	}
+
+	padLength := length - len(data)
+	paddedData := make([]byte, length)
+	for i := range paddedData {
+		paddedData[i] = 0
+		if i >= padLength {
+			paddedData[i] = data[i-padLength]
+		}
+	}
+	return newBuffer(paddedData)
+}
+
 func (b *byteBuffer) MarshalJSON() ([]byte, error) {
 	return json.Marshal(b.base64())
 }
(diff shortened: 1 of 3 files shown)


csstaub:
Hi!

Thanks for the pull request. I will take a closer look at this shortly.

First things first though:
* Have you signed our [Individual Contributor License Agreement](https://docs.google.com/a/squareup.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?ndplr=1&formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ) yet?
* Looks like you broke the build for Go 1.2/1.3, can you see if we can fix that? It wouldn't be the end of the world to drop support for old versions, but we should try to stay compatible if we can.

encoding.go
```diff
@@ -131,6 +131,22 @@ func newBuffer(data []byte) *byteBuffer {
...
+
+	padLength := length - len(data)
+	paddedData := make([]byte, length)
+	for i := range paddedData {
+		paddedData[i] = 0
+		if i >= padLength {
+			paddedData[i] = data[i-padLength]
+		}
+	}
+	return newBuffer(paddedData)
```
csstaub: Explicit array initialization shouldn't be necessary, see http://blog.golang.org/go-slices-usage-and-internals.

I believe the following block of code is sufficient:

	if len(data) >= length {
		return newBuffer(data)
	}
	pad := make([]byte, length - len(data))
	return newBuffer(append(pad, data...))

csstaub: See also #46

bifurcation:
Sorry for the delay on this, @csstaub.  It looks like boulder will now have a real requirement for thumbprints, so I'm picking this back up.

I have signed the contributor agreement, updated the PR to use `newFixedSizeBuffer`, and hopefully things will now build with Go 1.2/1.3.

csstaub:
Hi @bifurcation, thanks for cleaning this up! Looks good to me.

PR #38, opened 2015-06-29
Ensure EC coordinates have fixed length

bifurcation:
The [JWA specification for EC keys](http://tools.ietf.org/html/rfc7518#section-6.2.1.2) requires that the coordinates have a length fixed by the curve:

> The length of this octet string [representing the coordinate] MUST
> be the full size of a coordinate for the curve specified in the "crv"
> parameter.  For example, if the value of "crv" is "P-521", the octet
> string must be 66 octets long.

The JWK serialization code needs to render the coordinates into a fixed-size buffer, rather than only using as many bytes as it takes to represent the number.

(NB: This PR is based on #37; the only new stuff is in jwk.go:fromEcPublicKey)

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -131,6 +131,22 @@ func newBuffer(data []byte) *byteBuffer {
 	}
 }
 
+func newZeroPaddedBuffer(data []byte, length int) *byteBuffer {
+	if len(data) >= length {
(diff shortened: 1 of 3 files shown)


csstaub:
Obsoleted by #46

Issue #40, opened 2015-07-23
Invalid JWK exponent encoding for RSA ?


r0ro:
I believe there is an issue regarding exponent ("e") encoding in JWK.
In [jwk_test.go] (https://github.com/square/go-jose/blob/master/jwk_test.go#L181) the TestMarshalNonPointer JWK encoding test looks strange to me :

```go
	keyJson := []byte(`{
		"e": "AQAB",
		"kty": "RSA",
		"n": "vd7rZIoTLEe-z1_8G1FcXSw9CQFEJgV4g9V277sER7yx5Qjz_Pkf2YVth6wwwFJEmzc0hoKY-MMYFNwBE4hQHw"
	}`)

[ ... ]

expected := "{\"Key\":{\"kty\":\"RSA\",\"n\":\"vd7rZIoTLEe-z1_8G1FcXSw9CQFEJgV4g9V277sER7yx5Qjz_Pkf2YVth6wwwFJEmzc0hoKY-MMYFNwBE4hQHw\",\"e\":\"AAEAAQ\"}}"

```

Why isn't "e" encoded as "AQAB" as in input ?
as per [JWA RFC7518] (https://tools.ietf.org/html/rfc7518#section-6.3.1.2) e should be encoded in Base64urlUInt
AQAB represent 0x010001 => 65537
AAEAAQ represent 0x01000100 => 16777472

csstaub:
It's big-endian, so it decodes to the same number: `AQAB` decodes to three bytes (01 00 01) and `AAEAAQ` decodes to four bytes (00 01 00 01). But you are right, there is some extraneous zero padding there that we should probably get rid of for cleanliness.

r0ro:
yes you are right, I got the endianess wrong when printing. The issue is not trimming the extra 0.
I've tested with a public key with a 3 exponent :

```
-----BEGIN PUBLIC KEY-----
MIIBIDANBgkqhkiG9w0BAQEFAAOCAQ0AMIIBCAKCAQEAspbGSo1vhexIC/bTdYDW
eUodxH5VFBfOvtaRk650giucOFbAM6ZUZXalWxPJn/6XmXmxlTk6GKMJmAUdG/Gr
LxzSTvhbYWSRV1ViYtpJv7FTfH2oY038jtp33tSoXa0wdryxh3iVwct6DcE+WzVJ
[...]

PR #41, opened 2015-07-23
Fixes issue #40: extraneous zero padding on serialized exponent

csstaub:
@dgalling @r0ro

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -20,6 +20,7 @@ import (
 	"bytes"
 	"compress/flate"
 	"encoding/base64"
+	"encoding/binary"
 	"encoding/json"
 	"io"
 	"math/big"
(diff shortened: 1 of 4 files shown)


encoding.go
```diff
@@ -175,3 +184,15 @@ func (b byteBuffer) bigInt() *big.Int {
 func (b byteBuffer) toInt() int {
 	return int(b.bigInt().Int64())
 }
+
+func trimZeros(data []byte) []byte {
```
dgalling: Could you use `bytes.TrimLeft` instead?

csstaub: Derp, yes

encoding.go
```diff
@@ -131,6 +132,14 @@ func newBuffer(data []byte) *byteBuffer {
 	}
 }
 
+func newBufferFromInt(num uint64) *byteBuffer {
+	data := make([]byte, 8)
+	binary.BigEndian.PutUint64(data, num)
+	return &byteBuffer{
```
dgalling: Minor: I prefer we `return newBuffer(bytes.TrimLeft(data, "\x00"))`, but it's up to you.

csstaub: Yep, I like that better too. Will update and merge after lunch

PR #43, opened 2015-07-29
Add JSON marshal/unmarshal for JWS

bifurcation:


bifurcation:
Closing this, as (1) it's no longer a requirement for boulder and (2) it needs to have the nonce stuff factored out.

PR #44, opened 2015-07-30
Avoid panic when protected == nil

bifurcation:
Currently, the library panics when it is asked to serialize a JWS object with no protected header.  Instead, it should just not attempt to serialize the protected header, and the JSON serializer will then omit the corresponding field.


bifurcation:
Oops, meant this to be against the letsencrypt fork.

PR #46, opened 2015-08-18
Use fixed size buffers for EC coordinates

csstaub:
Coordinates must have a fixed size.

Changes:
  * Add new constructor for fixed-size byte buffers
  * Add helper functions for curve name/size
  * Update JWK serialization code

R: @dgalling @sqshh

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -132,6 +132,14 @@ func newBuffer(data []byte) *byteBuffer {
 	}
 }
 
+func newFixedSizeBuffer(data []byte, length int) *byteBuffer {
+	if len(data) > length {
+		panic("square/go-jose: invalid call to newFixedSizeBuffer (len(data) > length)")
+	}
+	pad := make([]byte, length-len(data))
+	return newBuffer(append(pad, data...))
+}
+
 func newBufferFromInt(num uint64) *byteBuffer {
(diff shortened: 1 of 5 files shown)


shared.go
```diff
@@ -182,3 +184,31 @@ func (dst *rawHeader) merge(src *rawHeader) {
...
+	case elliptic.P384():
+		return "P-384", nil
+	case elliptic.P521():
+		return "P-521", nil
+	default:
+		return "", fmt.Errorf("square/go-jose: unsupported/unknown elliptic curve")
+	}
+}
+
+// Get size of curve
```
dgalling: in bytes?

csstaub: fixed

encoding.go
```diff
@@ -132,6 +132,14 @@ func newBuffer(data []byte) *byteBuffer {
 	}
 }
 
+func newFixedSizeBuffer(data []byte, length int) *byteBuffer {
```
alokmenghrajani: comment saying the buffer is prepended with 0's?

alokmenghrajani:
lgtm

csstaub:
(Sidenote: for reference on the change, see http://www.rfc-editor.org/rfc/rfc7518.txt section 6.2.1.2)

PR #47, opened 2015-08-18
Fix flaky CBC-HMAC tests

csstaub:
Fix flaky CBC-HMAC tests. Previously, the `TestInvalidCiphertext` unit test would sometimes fail because the decrypted ciphertext could (per chance) still have valid padding. Also fixes an issue where `Open` would modify the ciphertext buffer in-place.

R: @dgalling @alokmenghrajani @sqshh

### DIFF
--- a/cipher/cbc_hmac.go
+++ b/cipher/cbc_hmac.go
@@ -112,7 +112,14 @@ func (ctx *cbcAEAD) Open(dst, nonce, ciphertext, data []byte) ([]byte, error) {
 	}
 
 	cbc := cipher.NewCBCDecrypter(ctx.blockCipher, nonce)
-	buffer := []byte(ciphertext[:offset])
+
+	// Make copy of ciphertext buffer, don't want to modify in place
+	buffer := append([]byte{}, []byte(ciphertext[:offset])...)
+
+	if len(buffer)%ctx.blockCipher.BlockSize() > 0 {
+		return nil, errors.New("square/go-jose: invalid ciphertext (invalid length)")
+	}
+
 	cbc.CryptBlocks(buffer, buffer)
 
 	// Remove padding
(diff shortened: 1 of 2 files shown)


cipher/cbc_hmac.go
```diff
@@ -112,7 +112,14 @@ func (ctx *cbcAEAD) Open(dst, nonce, ciphertext, data []byte) ([]byte, error) {
 	}
 
 	cbc := cipher.NewCBCDecrypter(ctx.blockCipher, nonce)
-	buffer := []byte(ciphertext[:offset])
+
+	// Make copy of ciphertext buffer, don't want to modify in place
+	buffer := append([]byte{}, []byte(ciphertext[:offset])...)
+
+	if len(buffer)%ctx.blockCipher.BlockSize() > 0 {
```
sqshh: Space between arguments in modulo?

sqshh:
LGTM.

cipher/cbc_hmac.go
```diff
@@ -112,7 +112,14 @@ func (ctx *cbcAEAD) Open(dst, nonce, ciphertext, data []byte) ([]byte, error) {
 	}
 
 	cbc := cipher.NewCBCDecrypter(ctx.blockCipher, nonce)
-	buffer := []byte(ciphertext[:offset])
+
+	// Make copy of ciphertext buffer, don't want to modify in place
+	buffer := append([]byte{}, []byte(ciphertext[:offset])...)
```
dgalling: `copy` might be a more natural function to use here

cipher/cbc_hmac.go
csstaub: It's formatted like this automatically by `go fmt`.

cipher/cbc_hmac.go
csstaub: If I use `copy` I will have to create a new byte array, then make the copy. This does both in one.

PR #48, opened 2015-08-18
Add additional test cases

csstaub:
This adds some additional test cases for stuff that wasn't covered before.

R: @alokmenghrajani @dgalling @sqshh

### DIFF
--- a/jwk_test.go
+++ b/jwk_test.go
@@ -221,6 +221,9 @@ func TestMarshalNonPointer(t *testing.T) {
 }
 
 func TestMarshalUnmarshalInvalid(t *testing.T) {
+	invalidCoord := make([]byte, curveSize(ecTestKey256.Curve)+1)
+	invalidCoord[0] = 1
+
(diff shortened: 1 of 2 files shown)


jwk_test.go
```diff
@@ -221,6 +221,9 @@ func TestMarshalNonPointer(t *testing.T) {
 }
 
 func TestMarshalUnmarshalInvalid(t *testing.T) {
+	invalidCoord := make([]byte, curveSize(ecTestKey256.Curve)+1)
+	invalidCoord[0] = 1
```
sqshh: Is there a significance to setting the first byte to 1?

csstaub: Yeah, I'm trying to generate a big integer that is too large for the curve. That's why I'm making a new byte array with `curveSize + 1`, and then setting the first byte to `1` (because otherwise it's just zero).

csstaub: Added a comment to clarify.

PR #51, opened 2015-10-08
Add support for the nonce header parameter

bifurcation:
ACME defines a 'nonce' header parameter for JWS, which must only be sent in a protected header.  This PR makes the following changes:

* A `Nonce` field to `JoseHeader` to expose nonce values from parsed objects
* A `AddNonces` method to signers to enable sending of nonces
* Checks for unprotected nonces in parsing code
* Tests that verify that the nonces are parsed and checked correctly

### DIFF
--- a/jws.go
+++ b/jws.go
@@ -70,8 +70,8 @@ func ParseSigned(input string) (*JsonWebSignature, error) {
 // Get a header value
 func (sig Signature) mergedHeaders() rawHeader {
 	out := rawHeader{}
-	out.merge(sig.protected)
-	out.merge(sig.header)
+	out.merge(sig.protected, true)
+	out.merge(sig.header, false)
 	return out
 }
 
--- a/jws.go
+++ b/jws.go
@@ -70,8 +70,8 @@ func ParseSigned(input string) (*JsonWebSignature, error) {
 // Get a header value
 func (sig Signature) mergedHeaders() rawHeader {
 	out := rawHeader{}
-	out.merge(sig.protected, true)
-	out.merge(sig.header, false)
+	out.merge(sig.protected)
+	out.merge(sig.header)
 	return out
 }
 
@@ -125,6 +125,10 @@ func (parsed *rawJsonWebSignature) sanitized() (*JsonWebSignature, error) {
 			}
 		}
 
+		if parsed.Header != nil && parsed.Header.Nonce != "" {
+			return nil, fmt.Errorf("square/go-jose: Nonce parameter included in unprotected header")
+		}
+
 		signature.header = parsed.Header
 		signature.Signature = parsed.Signature.bytes()
 		// Make a fake "original" rawSignatureInfo to store the unprocessed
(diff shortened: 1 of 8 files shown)


bifurcation:
@csstaub I think this is the last change we need to swap boulder back to using this repo (as opposed to the letsencrypt fork).  Your attention would be appreciated!

jws.go
```diff
@@ -155,6 +159,11 @@ func (parsed *rawJsonWebSignature) sanitized() (*JsonWebSignature, error) {
 			}
 		}
 
+		// Check that there is not a nonce in the unprotected header
+		if sig.Header != nil && sig.Header.Nonce != "" {
+			return nil, fmt.Errorf("square/go-jose: Nonce parameter included in unprotected header")
```
csstaub: Since this error gets reused everywhere, I would prefer if it was defined in a (private) constant. That should also make it easier to match on the specific error for testing.

jws_test.go
```diff
@@ -90,6 +90,38 @@ func TestFullParseJWS(t *testing.T) {
...
+	_, err := ParseSigned(input)
+	if err == nil {
+		t.Error("JWS with an unprotected nonce parsed as valid.")
+	} else if err.Error() != "square/go-jose: Nonce parameter included in unprotected header" {
+		t.Errorf("Improper error for unprotected nonce: %v", err)
+	}
+
+	// Full JSON
+	input = `{
+			"payload": "does-not-matter",
```
csstaub: Nit: whitespace

signing.go
```diff
@@ -156,6 +168,13 @@ func (ctx *genericSigner) Sign(payload []byte) (*JsonWebSignature, error) {
 	return obj, nil
 }
 
+// AddNonces provides or updates a nonce pool to the first recipients.
+// After this method is called, the signer will consume one nonce per
+// signature, returning an error if there are no nonces left in the pool.
+func (ctx *genericSigner) AddNonces(nonces []string) {
```
csstaub: Rather than having a nonce pool that gets exhausted, would it be preferable to simply pass the signer a callback function for retrieving nonces? E.g. something like `SetNonceProvider(func getNonce() string)` and have the signer call `getNonce` when one is required.

csstaub:
Looks good, the only real concern I have is the addition of `AddNonces` to the API. Changing the API in the future would be painful so I'd like to consider that a bit before we merge it. Do you have any particular design considerations/requirements for setting nonces?

bifurcation:
Thanks for the comments.  I'll fix the obvious ones. 

With regard to the API change: I admit that none of the options are super great here.  I think my preferred option would be to define an interface of the form:

```
type NonceSource interface {
  Nonce() (string, error)
}
```

It is a bit heavier-weight than just having a callback, but it's basically what we've done [in boulder](https://github.com/letsencrypt/boulder/blob/master/core/nonce.go#L96), so it would be easy to plug one into the other.

csstaub:
An interface works for me. I think that's a good, clean solution.

bifurcation:
@csstaub Think this is good to go now.

csstaub:
Looks good, thanks!

PR #53, opened 2015-10-16
Add support for JWK Sets

burrbd:
Nice package. Would you be open to adding support for [JWK Sets](https://tools.ietf.org/html/draft-ietf-jose-json-web-key-41#section-5)?

Thanks

### DIFF
--- a/jwk.go
+++ b/jwk.go
@@ -56,6 +56,67 @@ type JsonWebKey struct {
 	Algorithm string
 }
 
+// rawJsonWebKeySet represents a JWK Set with raw members.
+type rawJsonWebKeySet struct {
+	Keys []json.RawMessage `json:"keys"`
+}
+
+// JsonWebKeySet represents a JWK Set object.
+type JsonWebKeySet struct {
+	Keys []JsonWebKey
+}
+
+// MarshalJSON serializes the given set to its JSON representation.
+func (s JsonWebKeySet) MarshalJSON() ([]byte, error) {
+	var rawSet rawJsonWebKeySet
+
+	for _, key := range s.Keys {
+		data, err := key.MarshalJSON()
+		if err != nil {
+			return nil, err
+		}
+		rawSet.Keys = append(rawSet.Keys, data)
+	}
+
+	return json.Marshal(rawSet)
+}
+
+// UnmarshalJSON reads a set from its JSON representation.
+func (s *JsonWebKeySet) UnmarshalJSON(data []byte) error {
+	var rawSet rawJsonWebKeySet
+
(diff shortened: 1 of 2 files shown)


jwk.go
```diff
@@ -56,6 +56,67 @@ type JsonWebKey struct {
 	Algorithm string
 }
 
+// rawJsonWebKeySet represents a JWK Set with raw members.
+type rawJsonWebKeySet struct {
+	Keys []json.RawMessage `json:"keys"`
```
csstaub: If `JsonWebKey` implements `MarshalJSON` and `UnmarshalJSON`, would it not be possible to put `[]JsonWebKey` in here and have the `json` package parse it directly?

jwk_test.go
```diff
@@ -414,3 +414,45 @@ func TestThumbprint(t *testing.T) {
...
+	jsonbar2, err := set2.MarshalJSON()
+	if err != nil {
+		t.Errorf("problem marshalling set", err)
+	}
+	if !bytes.Equal(jsonbar, jsonbar2) {
+		t.Error("roundtrip should not lose information")
+	}
+}
+
+func TestJWKSetKey(t *testing.T) {
```
csstaub: Maybe we should add a test to make sure this holds true: ```The member names within a JWK Set MUST be unique; JWK Set parsers
   MUST either reject JWK Sets with duplicate member names or use a JSON
   parser that returns only the lexically last duplicate member name, as
   specified in Section 15.12 (The JSON Object) of ECMAScript 5.1
   [ECMAScript].```

jwk.go
```diff
@@ -56,6 +56,67 @@ type JsonWebKey struct {
...
+// UnmarshalJSON reads a set from its JSON representation.
+func (s *JsonWebKeySet) UnmarshalJSON(data []byte) error {
+	var rawSet rawJsonWebKeySet
+
+	if err := json.Unmarshal(data, &rawSet); err != nil {
+		return err
+	}
+
+	for _, raw := range rawSet.Keys {
+
```
csstaub: Nit: extra whitespace

csstaub:
Hi! Yes, we can certainly add support for JWK sets. I added some comments to your code, but in general it looks pretty good. You'll have to sign our CLA before we can accept, see https://github.com/square/go-jose/blob/master/CONTRIBUTING.md

jwk.go
burrbd: Wow cool. Wasn't aware of those interfaces. Makes code much simpler.

jwk_test.go
burrbd: Okay. I've added a test for this.

csstaub:
Looks good, thanks!

PR #55, opened 2015-11-02
add 'Use' element

### LINKED ISSUE #54: Consider adding "use" key for JWKs
I have seen that for some of the JWK-related libraries out there the "use" key in the JWK is required to find the correct key to use for sign verification and what not.

For example, JWK like:

```
{ keys: [ { "kty": "RSA", ... other elements } ] }
```

Fail, but this works

```
{ keys: [ { "use": "sig", "kty": "RSA", ... other elements } ] }
```

Now, for the stuff I'm doing, I want to generate the JWK file from the RSA private key on the fly. So I do something like the following:

```
type MyJWK struct {
   *jose.JsonWebKey
   Use string
}

json.Marshal(MyJWK{ Use: "sig", JsonWebKey: &jose.JsonWebKey{...} })
```

and for serialization/deserialization, I'm jumping through hoops, which is doable, but it sure would be nice if `jose.JsonWebKey` just did it for me. Thoughts?

lestrrat:
fixes #54

### DIFF
--- a/jwk.go
+++ b/jwk.go
@@ -115,7 +115,7 @@ func (k *JsonWebKey) UnmarshalJSON(data []byte) (err error) {
 	}
 
 	if err == nil {
-		*k = JsonWebKey{Key: key, KeyID: raw.Kid, Algorithm: raw.Alg}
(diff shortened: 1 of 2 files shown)


csstaub:
LGTM, thanks!

PR #57, opened 2015-11-12
Feature/single recipient kid support

lusis:
Okay I've gotten the tests written. I understand if you want to change/leave this out a bit but I think even with the possible new Recipient interface, a top-level Kid is still supported iirc so it's not wasted work.

### DIFF
--- a/jwe.go
+++ b/jwe.go
@@ -56,6 +56,15 @@ type recipientInfo struct {
 	encryptedKey []byte
 }
 
+// GetKid retrieces the (optional) kid attached to the object
+func (obj JsonWebEncryption) GetKid() string {
+	if obj.protected.Kid != "" || len(obj.protected.Kid) > 0 {
+		return obj.protected.Kid
+	}
+
+	return ""
+}
+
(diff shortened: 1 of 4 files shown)


jwe.go
```diff
@@ -56,6 +56,15 @@ type recipientInfo struct {
 	encryptedKey []byte
 }
 
+// GetKid retrieces the (optional) kid attached to the object
```
alokmenghrajani: s/retrieces/retrieves/

csstaub:
Sorry, but I don't think we should merge this to master in the current state. Due to the way Go does packaging I want to keep master as backwards-compatible as possible at all times, so we wouldn't want to merge an API that we don't want to support long-term. We should come up with a long-term solution for setting headers and then stick to that. I can however merge this to a non-master branch if you wish.

csstaub:
See https://github.com/square/go-jose/pull/58 for an alternate solution

zamicol:
Should this be rejected thanks to #58?

Issue #56, opened 2015-11-11
Custom headers and additional header support


lusis:
The JWE spec supports more than just the two fields allowed currently in go-jose (`alg` and `enc`).
[You can also specify a kid which is really important when using symmetric encryption like AES with JWE.](https://www.ietf.org/mail-archive/web/jose/current/msg04517.html)

I'm attempting something like so:

``` go
    alg := jose.KeyAlgorithm("dir")
    enc := jose.ContentEncryption("A256GCM")
    encrypter, err := jose.NewEncrypter(alg, enc, resp.Plaintext)
    if err != nil {
        log.Fatalf(err.Error())
    }
    obj, err := encrypter.Encrypt([]byte(contents))
    obj.Header.KeyID = "XXXXXX"
```

but when I attempt to base64 decode the header (first part), I still only see `alg` and `enc`.

I'm proposing either changing the Encrypter interface to support new funcs for adding `kid` specifically or a more generic one to allow setting headers that are unrelated to the encryption and key methods being used.

I can maybe make a PR on this but I need to know which way you'd rather go first.

Thanks!

csstaub:
Yeah, we currently don't have good support for that. I'm not sure what would be a good interface for doing this, e.g. how do you control whether you want to set headers for the entire message or only for individual recipients? I think it would make sense to change the signature of e.g. `NewEncrypter`, `AddRecipient` to allow for also passing in headers. Another option would be to wrap the key in a new interface and support taking that as input. For example, we could consider adding a new `Recipient` interface which would contain the key but also metadata that describes it (such as key id or thumbprints). In `NewEncrypter` we would then distinguish between raw keys and the new recipient interface.

csstaub:
I think adding a new `Recipient` interface that can hold keys and metadata together would be a cleaner solution, but I'd love to have some input from more people on this. Thoughts?

lusis:
I was just looking. Can we maybe as a first pass consider using the same mechanism as `SetCompression` as a short term fix? I can knock that out fairly quickly (it's what I was going to do in the short term as a fork anyway).

lusis:
Lemme also poke at the recipient interface as well. Obviously the whole `kid` bit doesn't play well with `jose-util` but it doesn't really support `dir` anyway. In our use case we'd be passing in an AWS KMS alias as the kid and grabbing the key ourselves.

lusis:
FWIW currently in our case we don't need this to per-recipient

csstaub:
[...]

PR #58, opened 2015-11-17
Support symmetric keys in JWK and support JWK for encrypters

csstaub:
r: @alokmenghrajani @mcpherrinm @sqshh 
cc: @lusis

This pull request adds:
* Support for representing symmetric keys as a JsonWebKey
* Support for instantiating an encrypter given a JsonWebKey

This partially addresses #56, in that it allows for creating encrypters with custom key IDs.

For example:
```
crypter, err := NewEncrypter(A128KW, A128GCM, &JsonWebKey{
		KeyID:   "test-id",
		Key:     []byte{0, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15},
	})
```
The above snippet would create an encrypter that adds the key ID "test-id" to any output message.

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,13 +67,15 @@ has a list of constants.
 
 See below for a table of supported key types. These are understood by the
 library, and can be passed to corresponding functions such as `NewEncrypter` or
-`NewSigner`.
+`NewSigner`. Note that if you are creating a new encrypter or signer with a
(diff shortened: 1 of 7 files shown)


alokmenghrajani:
lgtm

PR #59, opened 2015-11-17
Reconcile with letsencrypt/go-jose

bifurcation:
There's only one significant change that has been added to letsencrypt/go-jose that we need to push upstream: A fix to not panic when processing a JWE or JWS without a protected header.

(`go fmt` also got rid of an extra space in `jwk_test.go`)

### DIFF
--- a/jwe.go
+++ b/jwe.go
@@ -227,7 +227,8 @@ func parseEncryptedCompact(input string) (*JsonWebEncryption, error) {
 
 // CompactSerialize serializes an object using the compact serialization format.
 func (obj JsonWebEncryption) CompactSerialize() (string, error) {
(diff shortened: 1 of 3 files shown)


bifurcation:
There are a lot more commits here than you might expect, due to this PR being based on letsencrypt/go-jose.  Most of these are washed out in the merge commit a17ba26.

csstaub:
Looks good :+1:

csstaub:
Would you mind cleanup up the commit history a bit? Not all authors have signed our CLA, and since most of the commits turn out to be no-ops anyway it'll be easier to just squash the history.

bifurcation:
OK, should be squashed now, @csstaub

csstaub:
Great. Thanks!

PR #61, opened 2015-11-19
Fix error message in rawJsonWebSignature#sanitized

### LINKED ISSUE #60: Incorrect error in `sanitized()`
https://github.com/square/go-jose/blob/master/jws.go#L106
```
func (parsed *rawJsonWebSignature) sanitized() (*JsonWebSignature, error) {
	if parsed.Payload == nil {
		return nil, ErrUnprotectedNonce
	}
```

I think this was incorrectly changed from its old value ` fmt.Errorf("square/go-jose: missing payload in JWS message")`. @bifurcation

csstaub:
Fixes #60


alokmenghrajani:
:+1:

PR #63, opened 2016-02-06
Optionally disable embedding JWK in protected header

boivie:
In some applications, the signer's public key is already known to
the recipient, or the 'kid' property in the header can be used to
lookup the key. Simply speaking, embedding the signer's public key
in the protected header's 'jwk' property is sometimes not wanted.

This commit adds a new method to the Signing interface to disable
embedding of the public key of the signer. It will be embedded by
default to maintain backwards compatibility.

### DIFF
--- a/signing.go
+++ b/signing.go
@@ -31,12 +31,14 @@ type NonceSource interface {
 type Signer interface {
 	Sign(payload []byte) (*JsonWebSignature, error)
 	SetNonceSource(source NonceSource)
+	SetEmbedJwk(embed bool)
 }
 
 // MultiSigner represents a signer which supports multiple recipients.
 type MultiSigner interface {
 	Sign(payload []byte) (*JsonWebSignature, error)
 	SetNonceSource(source NonceSource)
+	SetEmbedJwk(embed bool)
 	AddRecipient(alg SignatureAlgorithm, signingKey interface{}) error
 }
 
@@ -51,6 +53,7 @@ type payloadVerifier interface {
 type genericSigner struct {
 	recipients  []recipientSigInfo
 	nonceSource NonceSource
+	embedJwk    bool
 }
 
 type recipientSigInfo struct {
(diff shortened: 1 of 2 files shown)


signing.go
```diff
@@ -31,12 +31,14 @@ type NonceSource interface {
...
 	Sign(payload []byte) (*JsonWebSignature, error)
 	SetNonceSource(source NonceSource)
+	EmbedJwk(embed bool)
 }
 
 // MultiSigner represents a signer which supports multiple recipients.
 type MultiSigner interface {
 	Sign(payload []byte) (*JsonWebSignature, error)
 	SetNonceSource(source NonceSource)
+	EmbedJwk(embed bool)
```
csstaub: Rename this to `SetEmbedJwk`

csstaub:
This looks good, but makes me feel we should probably refactor the Signer and Encrypter interfaces altogether. I think it would be preferable if signers/encrypters were immutable and options like this could only be set when building the initial object. I'll put this on my TODO list for a v2 maybe.

boivie:
I have updated my commit to use SetEmbedJwk instead.

Regarding immutable signers, I do agree. I would also like to see a way to add per-message unique fields (such as "cty") in the (protected) header, which would also warrant an interface update. I guess v2 is just on the drawing board?

csstaub:
Well, it's "on the drawing board" in the sense that "I don't like the way it is now and I'd like to clean it sometime". Also, per-message unique fields would be great! Another candidate for a v2 addition.

csstaub:
Ah, can you also sign our CLA? See here: https://docs.google.com/a/squareup.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1

boivie:
Certainly - I have signed the CLA

Thanks for merging the PR!

csstaub:
Thanks, and thanks for your contribution!

Issue #67, opened 2016-02-11
Registered claims


yawn:
Are you interested in supporting [registered claims](https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#rfc.section.4.1) in this project? Or should I create a separate repository implementing (at least a subset of those) on top of `go-jose`?

I'd like to support (at least) `exp` and `nbf`.

csstaub:
Yeah, I think this could make a good addition for go-jose.v2, if you're interested in contributing it.

yawn:
No chance to get in into the current version? I'd build it as an independent package.

csstaub:
Sure, we can merge it in anytime. I was just mentioning v2 because I'm planning to break the interface for that anyway, so if you feel interface-breaking change would be needed it could go in there.

csstaub:
See `jwt` sub-package in the v2 branch for a dev version.

PR #69, opened 2016-02-17
Added validation of registered claims

yawn:
See #67 for a discussion of the rationale.

### DIFF
--- a/claims/claims.go
+++ b/claims/claims.go
@@ -1,0 +1,95 @@
+package claims
+
+import (
+	"encoding/json"
+	"time"
+)
+
+const (
+
+	// Audience identifies the target audience, see https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#audDef
+	Audience = "aud"
+
+	// Expires identifies the time after which the token must not be accepted, see https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#expDef
+	Expires = "exp"
+
+	// IssuedAt identifies the time the token was issued at, see https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#iatDef
+	IssuedAt = "iat"
+
+	// Issuer identifies the principal who created the token, see https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#issDef
+	Issuer = "iss"
+
+	// JwtID identifies this token uniquely, see https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#jtiDef
+	JwtID = "jti"
+
+	// NotBefore identifies the time after which this token becomes valid, see https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#nbfDef
+	NotBefore = "nbf"
+
+	// Subject identifies the principal this token is intended for, see https://self-issued.info/docs/draft-ietf-oauth-json-web-token.html#subDef
+	Subject = "sub"
+)
+
+// Assertion verifies a given Claims value
+type Assertion interface {
+	Verify(claim Claim, value interface{}) error
+}
+
+type (
+	// Assertions maps a Claim to a number of matching Assertions
+	Assertions map[Claim][]Assertion
+
+	// Claim identifies a reserved (well-known) claim
+	Claim string
+
+	// Claims maps a Claim to it's value
+	Claims map[Claim]interface{}
+)
+
+// NewClaims initializes a serializable Claims structure
+func NewClaims() (claims Claims) {
+	claims = make(map[Claim]interface{})
+	return claims
+}
+
+// Bytes serializes the Claims to JSON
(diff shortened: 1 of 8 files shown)


yawn:
The tests depend on https://github.com/stretchr/testify and won't pass without it being installed or vendored. You should be able to test it locally after `go get`. Feel free to rip out the additional testing helpers if you don't like the extra dependency.

In the current configuration Travis did *NOT* test the package.

csstaub:
Thanks for your contribution! I will try to review this today.

First things first though, before we can merge this you will need to sign our contributor agreement here: https://github.com/square/go-jose/blob/master/CONTRIBUTING.md. Also, it would be nice to make travis-ci test the new package too.

claims/verifier.go
```diff
@@ -0,0 +1,83 @@
...
+func NewVerifier() *Verifier {
+
+	return &Verifier{
+		Assertions: make(map[Claim][]Assertion),
+	}
+
+}
+
+// Add adds an Assertion for a given Claim
+func (v *Verifier) Add(c Claim, a Assertion) *Verifier {
```
csstaub: Why not make the verifier immutable? Just take a list of assertions about claims in `NewVerifier`, that way you can't accidentally have a verifier where you forget to add assertions.

claims/claims.go
```diff
@@ -0,0 +1,95 @@
...
+func (c Claims) MarshalBinary() ([]byte, error) {
+	return json.Marshal(c)
+}
+
+// NotBefore adds an not before claim as UTC integer
+func (c Claims) NotBefore(t time.Time) Claims {
+	return c.Add(NotBefore, c.t2i(t))
+}
+
+func (c Claims) t2i(t time.Time) int {
```
csstaub: Use a more descriptive name here

claims/time.go
```diff
@@ -0,0 +1,66 @@
...
+type TimeSource func() time.Time
+
+func defaultTimeSource() time.Time {
+	return time.Now()
+}
+
+// Time handles time related registered claim verification, using an optional
+// TimeSource as point of reference
+type Time struct {
+	T TimeSource
```
csstaub: Also, use a more descriptive name here. Stuff like `t.T` from below is kinda confusing otherwise.

claims/time.go
```diff
@@ -0,0 +1,66 @@
...
+
+			if v > now {
+				return fmt.Errorf(errIat, v, now)
+			}
+
+		}
+
+	}
+
+	return nil
```
csstaub: What happens if this function is called with e.g. `Expires` as claim and `nil` as value? I feel like that should return an error to guard against type confusion. The switch statement above should have a default clause that catches invalid types and returns an error.

csstaub:
Looks pretty good! I added some comments, but nothing major. One more thing I'd like is to add license headers (with your name included); see the other source files for examples.

claims/verifier.go
yawn: Currently for supporting the [builder pattern](https://en.wikipedia.org/wiki/Builder_pattern), but there is no reason that could not be immutable. I could change it, I suppose.

claims/claims.go
yawn: `timeToInt`?

claims/time.go
yawn: Ok.

claims/time.go
yawn: This is handled by the `Verifier`. Have a look here: https://github.com/yawn/go-jose/blob/claims/claims/verifier.go#L65

yawn:
I can add the license headers and fix Travis. I'd prefer to keep using testify (the code has been exported from a private repository using `go-jose`) though. Would you mind if we get that in as a testing-only dependency? I find it incredibly handy and non-invasive.

csstaub:
Yeah, testify is great! No problems with including it whatsoever.

claims/verifier.go
csstaub: Don't feel strongly about it, so feel free to leave it for now. I noted it because I'm moving towards immutable encrypters/signers in the go-jose.v2 base package.

claims/claims.go
csstaub: Works for me

claims/time.go
csstaub: It's true that the verifier handles it, but since this is a public interface you don't know what values other users will be calling your function with.

yawn:
This will take a little while longer. Hopefully on the weekend.

yawn:
Just FYI - I have not forgotten this PR. I'll come back to it before the end of the month.

csstaub:
Sounds good :+1:

zamicol:
Should this now be rejected in favor of @shaxbee's additions? See PR #114

Issue #73, opened 2016-03-09
Go's JSON parser is case-insensitive


jsha:
As pointed out at https://www.ietf.org/mail-archive/web/json/current/msg03763.html, Go's JSON parser is case-insensitive, which can lead to differences in interpretation between go-jose and libraries in other languages.

One possible workaround, mentioned in that thread, is to parse into a `map[string]string`, where the key names will be based on exact match.

csstaub:
Parsing as a `map[string]string` would work, but it's super annoying to parse messages that way. Maybe we could just fork `encoding/json` and make a case-insensitive version? That way we could also make changes to reject JSON blobs with duplicate keys.

rolandshoemaker:
:+1: for forking `encoding/json`, none of these bugs (or... idk, quirks?) will be fixed during the go1 lifetime to retain backwards compatibility.

PR #74, opened 2016-03-09
Switch to case-sensitive encoding/json fork

csstaub:
r: @alokmenghrajani @mcpherrinm @sqshh @jsha @rolandshoemaker 

Fixes issue #73 by switching to a forked, case-sensitive version of `encoding/json` (from Go 1.6).

Changes in `square/go-jose`:
* Switch to forked `json` package
* Add tests for case-sensitive parsing behavior
* Drop Go 1.2 support (broken with forked `json`)

Changes in the forked `encoding/json` package:
* Remove `fold.go`, `fold_test.go` (contained case-sensitive folding functions)
* Set `equalFold` to be `bytes.Equal` in `encode.go` (no more case-sensitive matching)
* Removed `example_test.go` (broke build on older Go versions)

Also updated some of the test cases in the forked `json` package, as they were depending on the case-insensitive parsing behavior. In particular, `decode_test.go` (c.f `Ambig` struct) was updated.


alokmenghrajani:
:+1:

sqshh:
This mostly looks fine to me, but it is really hard to see what differs between old and new json here.

It may also be worth a note somewhere (probably the README) that we don't use Go's default json library for the case sensitivity issues.

csstaub:
Here's a diff from the original and the forked `json` packages:

```
cs go-jose case-insensitive-json $ diff -u json ~/src/golang/src/encoding/json
Only in json: LICENSE
Only in json: README.md
diff -u json/decode_test.go /Users/cs/src/golang/src/encoding/json/decode_test.go
--- json/decode_test.go	2016-03-09 14:29:29.000000000 -0800
+++ /Users/cs/src/golang/src/encoding/json/decode_test.go	2015-12-16 16:27:05.000000000 -0800
@@ -211,10 +211,9 @@
 }

 type Ambig struct {
-	// Given "hello", the second match should win.
+	// Given "hello", the first match should win.
 	First  int `json:"HELLO"`
-	Second int `json:"hello"`
-	Third  int `json:"Hello"`
+	Second int `json:"Hello"`
 }

 type XYZ struct {
@@ -376,7 +375,7 @@
 	{
 		in:  `{"hello": 1}`,
 		ptr: new(Ambig),
-		out: Ambig{Second: 1},
+		out: Ambig{First: 1},
 	},

 	{
diff -u json/encode.go /Users/cs/src/golang/src/encoding/json/encode.go
--- json/encode.go	2016-03-09 14:29:29.000000000 -0800
+++ /Users/cs/src/golang/src/encoding/json/encode.go	2016-03-09 13:59:54.000000000 -0800
@@ -951,7 +951,7 @@

 func fillField(f field) field {
 	f.nameBytes = []byte(f.name)
-	f.equalFold = bytes.Equal
+	f.equalFold = foldFunc(f.nameBytes)
 	return f
 }

Only in /Users/cs/src/golang/src/encoding/json: example_test.go
Only in /Users/cs/src/golang/src/encoding/json: fold.go
Only in /Users/cs/src/golang/src/encoding/json: fold_test.go
diff -u json/stream_test.go /Users/cs/src/golang/src/encoding/json/stream_test.go
--- json/stream_test.go	2016-03-09 14:29:29.000000000 -0800
+++ /Users/cs/src/golang/src/encoding/json/stream_test.go	2016-03-09 13:59:54.000000000 -0800
@@ -334,7 +334,7 @@
 	defer res.Body.Close()

 	foo := struct {
-		Foo string `json:"foo"`
+		Foo string
 	}{}

 	d := NewDecoder(res.Body)
Common subdirectories: json/testdata and /Users/cs/src/golang/src/encoding/json/testdata
```

rolandshoemaker:
For Boulders use it'd be very nice to have a build flag included in `go-jose` which allows us to disable/enable this behavior (either way works, explicit disable might be better) as there will be a period where we will need to let users know about this upcoming change and we'd like to stay in-sync with any changes you guys make during that period.

csstaub:
@rolandshoemaker That seems reasonable, but I don't know of a good way to do conditional imports in Go. However it would be more or less sufficient to just replace the `github.com/square/go-jose/json` imports with `encoding/json`. Does that work for you? Otherwise let me know what you prefer.

sqshh:
LGTM

rolandshoemaker:
@csstaub Altering our own fork is totally an option, although we've been explicitly trying to get back to using upstream since it makes tracking actual changes/pushing fixes back much harder.

If you're willing to do it here I think the quickest solution would be to make a shim around `json.Marshal` and `json.Unmarshal` that references either your fork or the std lib along the lines of:

```
// +build std_json

package jose

import (
  "encoding/json
)

func MarshalJSON(v interface{}) ([]byte, error) { return json.Marshal(v) }
func UnmarshalJSON(data []byte, v interface{}) error { return json.Unmarshal(data, v) }
```

and

```
// +build !std_json

package jose

import (
  "github.com/square/go-jose/json
)

func MarshalJSON(v interface{}) ([]byte, error) { return json.Marshal(v) }
func UnmarshalJSON(data []byte, v interface{}) error { return json.Unmarshal(data, v) }
```

csstaub:
Ah, wasn't thinking about build tags! Yes, that is totally reasonable. We can implement that here.

csstaub:
@rolandshoemaker Added support for `std_json` build tag now. Please take a look.

rolandshoemaker:
LGTM, thanks!

PR #76, opened 2016-03-12
Reject duplicate keys in encoding/json fork

csstaub:
r: @alokmenghrajani @mcpherrinm @sqshh 

Reject duplicate keys in JSON objects. Also rm some unused code bits.

### DIFF
--- a/json/decode.go
+++ b/json/decode.go
@@ -601,6 +601,7 @@ func (d *decodeState) object(v reflect.Value) {
 	}
 
 	var mapElem reflect.Value
+	keys := map[string]bool{}
 
 	for {
 		// Read opening " of string key or closing }.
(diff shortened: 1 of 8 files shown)


csstaub:
Review, anyone?

json/decode.go
```diff
@@ -617,11 +618,19 @@ func (d *decodeState) object(v reflect.Value) {
 		start := d.off - 1
 		op = d.scanWhile(scanContinue)
 		item := d.data[start : d.off-1]
-		key, ok := unquoteBytes(item)
+		key, ok := unquote(item)
```
mcpherrinm: I'm not sure what's going on here.  Why are switching from bytes to string here?

mcpherrinm:
lgtm besides one point I'm unclear on.

json/decode.go
csstaub: It was previously calling `unquoteBytes` so it could do case-folding (by calling `equalFold` below), but that's not relevant anymore as I deleted all the case-folding logic. Calling `unquote` is equivalent to `unquoteBytes`, except that the return value is a string instead of the raw bytes of the key. Changed it because maps don't allow byte slice indices (because of their mutability I believe).

mcpherrinm: Sounds good to me. :+1: :shipit:

alokmenghrajani:
:+1:

PR #82, opened 2016-03-23
Claims encoding/decoding

shaxbee:
Claims represented as concrete type.

Follows JWT spec on interpretation of claim fields values, namely:
* NumericDate can be represented as either int or float
* Audience can be either string or array of strings

Timestamps are normalized by checking Time.IsZero, otherwise negative timestamp values would show up in encoded JSON.

Validator is checking Expiry/NotBefore using leeway as indicated in a spec.
Current time is provided as argument to Validator.Validate call.
Unit tests for validator are still pending.


shaxbee:
I'm gonna work on Validator tomorrow, next on is work on JWT interface to bring JWS and Validator under one roof.

jwt/json.go
```diff
@@ -0,0 +1,110 @@
...
+
+import (
+	"bytes"
+	"encoding/json"
+	"reflect"
+	"strconv"
+	"time"
+)
+
+type StringOrArray []string
```
shaxbee: This could be made private.

csstaub:
Thanks for your contribution!

First things first: Before we can accept code contributions, Square requires that contributors sign our [Individual Contributor License Agreement][1].  The files you added also need a copyright header (see other files in the repo). You can put your name in the copyright header if you wish. 

Other than that, the `go-jose` package is using a forked version of `encoding/json`, any new sub-packages should also use the fork. Ideally, this should also support using the `std_json` build tag do switch between the fork and the built-in version. Let me know if you have questions about this.

 [1]: https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1

shaxbee:
Signed the CLA. I'll follow up with headers / json package soon.

csstaub:
FYI: This currently conflicts with PR #69, and I haven't really reviewed either in enough detail to decide which one I'd rather merge yet, but I don't think it makes sense to merge both.

shaxbee:
License headers and jose JSON package usage added.

shaxbee:
Sample usage:
```go

token, err := jwt.ParseSigned(s, key)
if err != nil {
  panic(err)
}
err := token.Claims.Validate(Expected{Issuer: "valid-issuer", Time: time.Now})
if err != nil {
  panic(err)
}

priv := struct {
  Scopes []string
}{}
token.ParsePrivate(&priv)

fmt.Println("Scopes: ", priv.Scopes)
```

shaxbee:
BTW I'm considering two options for creating token from existing serialized source:
* ParseSigned(s string, key interface{}) / ParseEncrypted(s string, key interface{}) factory functions
* FromSigned(s *JsonWebSignature)
* FromEncrypted(e *JsonWebEncryption)

json_std_test.go
```diff
@@ -88,7 +88,7 @@ var JWKSetDuplicates = stripWhitespace(`{
 
 func TestDuplicateJWKSetMembersIgnored(t *testing.T) {
 	type CustomSet struct {
-		JsonWebKeySet
+		JSONWebKeySet
```
csstaub: Whoops, yeah, this changed on the `v2` branch. Thanks for catching that.

shaxbee:
Added custom claims support - claims can be provided in following forms:

```go
// Public claims only
c1 := &jwt.Claims{}
// parse claims
err := tok.Claims(c1)
// create token with claims
t, err := jwt.New(c1) 

// Private claims only
c2 := &struct {
  Scopes []string
}{}

// Public and private claims
c3 := &struct{
  jwt.Claims
  Scopes []string
}{}

// Map of public claims
c4 := map[string]interface{}{}
```

shaxbee:
@csstaub Any feedback on PR?

csstaub:
Sorry for the delay, I will get back to this soon.

jwt/builder.go
```diff
@@ -0,0 +1,128 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package jwt
+
+import "github.com/square/go-jose"
```
csstaub: Since this is on the `v2` branch, it should import `gopkg.in/square/go-jose.v2`.

jwt/builder.go
```diff
@@ -0,0 +1,128 @@
+/*-
+ * Copyright 2014 Square Inc.
```
csstaub: You can also add your name here if you want since you contributed this entire file (applies to other files too).

jwt/oauth2.go
```diff
@@ -0,0 +1,67 @@
...
+
+	raw := r.PostFormValue("assertion")
+	if raw == "" {
+		return nil, ErrNoToken
+	}
+
+	return p(raw)
+}
+
+// ToResponse writes token to standard OAuth 2.0 response
```
csstaub: Honestly I don't think this belongs in here, I'd like to keep this library focused on JOSE standards and OAuth is definitely outside of the scope of that.

jwt/json.go
```diff
@@ -0,0 +1,183 @@
...
+
+	c := f.Addr().Interface().(*Claims)
+	return c, nil
+}
+
+func marshalClaims(cl interface{}) ([]byte, error) {
+	switch cl := cl.(type) {
+	case *Claims:
+		return cl.marshalJSON()
+	case map[string]interface{}:
```
csstaub: Why support both types? I'd prefer to only unmarshal this into `*Claims` to provide type-safety.

jwt/builder.go
```diff
@@ -0,0 +1,128 @@
...
+			}
+			return e, e.Decrypt, nil
+		},
+	}
+}
+
+// Claims encodes claims into JWE/JWS form
+func (b *Builder) Claims(c interface{}) *Builder {
+	if b.transform == nil {
+		panic("Signer/Encrypter not set")
```
csstaub: This check is unnecessary: nil pointer checks happen implicitly on every deref, for free.

jwt/builder.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	if b.payload != nil {
+		panic("Claims already set")
```
csstaub: This seems fine to omit: on multiple calls to `Claims`, just overwrite the existing claims (or, perhaps, merge them?).

jwt/claims.go
```diff
@@ -0,0 +1,90 @@
...
+
+import (
+	"time"
+
+	"github.com/square/go-jose"
+)
+
+// Claims represents public claim values defined in RFC7519
+type Claims struct {
+	Issuer    string    `json:"-"`
```
csstaub: Why add `json:"-"` here?

jwt/claims.go
```diff
@@ -0,0 +1,90 @@
...
+	Iss string      `json:"iss,omitempty"`
+	Sub string      `json:"sub,omitempty"`
+	Aud audience    `json:"aud,omitempty"`
+	Exp NumericDate `json:"exp,omitempty"`
+	Nbf NumericDate `json:"nbf,omitempty"`
+	Iat NumericDate `json:"iat,omitempty"`
+	Jti string      `json:"jti,omitempty"`
+}
+
+func (c *Claims) marshalJSON() ([]byte, error) {
```
csstaub: You can make this `MarshalJSON` and `UnmarshalJSON` below to make it directly serializeable/deserializeable.

csstaub:
I left some comments, but I like the overall structure. It might be getting a bit too much to address this all in a single PR. Since we're merging this into the development `v2` branch, I think we can merge this in and keep iterating on it in smaller sets of changes. The only thing I'd like to have addressed before merging is to have the import paths fixed and removing the oauth functionality (I'm not convinced it should be part of this package).

jwt/claims.go
shaxbee: That would break cases when user-defined struct contains public claims.

shaxbee:
Done.

jwt/builder.go
shaxbee: Yes, i intend to rework claims setting so the claims are merged (eg. can be passed around), also the methods for setting public claims (for which we defined structure) will be separate from private claims, thus avoiding all hacks that are there right now.

jwt/json.go
shaxbee: That's current mechanism to support private (user-defined) claims.

jwt/oauth2.go
shaxbee: Removed :-)

jwt/builder.go
shaxbee: Done.

jwt/builder.go
shaxbee: Done.

jwt/builder.go
shaxbee: I did it for readability purposes, but have no problem removing it.

csstaub:
Great, thanks for your contribution! Merging this into `v2` branch for dev. I have lots of things I want to address in `v2` before it's ready for release though, which might include some more interface changes. So, details are still subject to change.

Issue #85, opened 2016-03-28
Support KeyID of symmetric keys in JWS


b6g:
I got a panic when I created a HS256 signer with a JsonWebKey, 

```
sr, err := jose.NewSigner(jose.HS256, &jose.JsonWebKey{
	Key:   key,
	KeyID: "k0",
})

sr.Sign([]byte("hello")
```

and the panic was,

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal 0xb code=0x1 addr=0x0 pc=0x46ecd0]

goroutine 1 [running]:
panic(0x5f1ec0, 0xc82000e150)
	/opt/go/src/runtime/panic.go:464 +0x3e6
github.com/square/go-jose.makeJWSRecipient(0x62f448, 0x5, 0x6014a0, 0xc820010300, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/b6g/git/src/github.com/square/go-jose/signing.go:131 +0x450
github.com/square/go-jose.(*genericSigner).AddRecipient(0xc820012780, 0x62f448, 0x5, 0x6014a0, 0xc820010300, 0x0, 0x0)
	/home/b6g/git/src/github.com/square/go-jose/signing.go:109 +0x69
github.com/square/go-jose.NewSigner(0x62f448, 0x5, 0x6014a0, 0xc820010300, 0x0, 0x0, 0x0, 0x0)
	/home/b6g/git/src/github.com/square/go-jose/signing.go:70 +0x167
main.sign(0xc82000a310, 0x1, 0x1)
```

The problem is `makeSymmetricSigner` does not allocate the `publicKey` field, yet `makeJWSRecipient` assigns the `KeyID` of the `signingKey` to `recipient.publicKey.KeyID`.

I could be wrong, but I think the spec allows setting "kid" when HS256 is used.

I could not find a good fix for this problem. If we new `publicKey` in `makeSymmetricSigner`, `genericSigner.Sign` will fail with json marshal error. This is because `ctx.embedJwk` is true and `protected.Jwk` will be set, but the `Key` field of the `JsonWebKey` is nil. I could work around this by calling `SetEmbedJwk(false)`, but it does not sound correct to me.

Thanks,

csstaub:
Yes, seems like a definite bug. Thanks for reporting this. I will take a look into fixing this later today.

PR #86, opened 2016-03-29
Allow setting kid header for symmetric signers

csstaub:
r: @alokmenghrajani @mcpherrinm @sqshh @b6g
Fixes issue #85. Signers previously incorrectly assumed that the key id would be embedded in the public key, but that is not always true, in particular for signers instantiated with symmetric keys. To fix this, create a new field on recipientSigInfo to store the key id instead.

Merging this into master, later going to cherry-pick separately into `v1` and tag a new version.

### DIFF
--- a/signing.go
+++ b/signing.go
@@ -58,6 +58,7 @@ type genericSigner struct {
 
 type recipientSigInfo struct {
 	sigAlg    SignatureAlgorithm
+	keyID     string
 	publicKey *JsonWebKey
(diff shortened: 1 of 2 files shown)


b6g:
LGTM, when can you merge it? Thanks,

csstaub:
Merged into `v1`, tagged as `v1.0.1` (https://github.com/square/go-jose/tree/v1.0.1)

PR #90, opened 2016-04-24
Add unit tests for unicode handling

csstaub:
@alokmenghrajani @mcpherrinm @sqshh 
See https://tools.ietf.org/html/rfc7515#section-10.13


mcpherrinm:
lgtm

PR #91, opened 2016-05-02
Add JsonWebKey.Valid method

rolandshoemaker:
Adds the method `Valid` to `JsonWebKey` to allow testing the validity of the wrapped key. It may make sense to utilize this in the Marshal/Unmarshal methods as well.

Unblocks letsencrypt/boulder#1728.

cc @jsha


csstaub:
LGTM. Looks like there's an unrelated build error; I can fix that and merge this later today.

rolandshoemaker:
:+1:

PR #92, opened 2016-05-02
Migrate jose-util to kingpin package

csstaub:
r: @mcpherrinm @alokmenghrajani @sqshh 
Migrate jose-util to alecthomas/kingpin package, instead of codegansta/cli. This brings it in line with the rest of our packages (e.g. we use kingpin in ghostunnel, keywhiz-fs). Also unblocks merging of #91.


mcpherrinm:
lgtm

PR #96, opened 2016-06-08
Support for x5c

### LINKED ISSUE #95: x5c / x5t parameters
Is it possible to use [x5c](https://tools.ietf.org/html/rfc7517#section-4.7) parameters with go-jose? I want to marshall a RSA private key along a x509 certificate (for HTTP over TLS).

Example: https://tools.ietf.org/html/rfc7517#appendix-B

arekkas:
This is a first draft to resolve #95 

ping @csstaub

### DIFF
--- a/crypter_test.go
+++ b/crypter_test.go
@@ -25,6 +25,10 @@ import (
 	"fmt"
 	"io"
 	"testing"
+	"crypto/x509/pkix"
+	"time"
+	"crypto/x509"
+	"math/big"
 )
 
 // We generate only a single RSA and EC key for testing, speeds up tests.
@@ -33,6 +37,26 @@ var rsaTestKey, _ = rsa.GenerateKey(rand.Reader, 2048)
 var ecTestKey256, _ = ecdsa.GenerateKey(elliptic.P256(), rand.Reader)
 var ecTestKey384, _ = ecdsa.GenerateKey(elliptic.P384(), rand.Reader)
 var ecTestKey521, _ = ecdsa.GenerateKey(elliptic.P521(), rand.Reader)
+var x509TestCertificate = func() *x509.Certificate {
+	serialNumberLimit := new(big.Int).Lsh(big.NewInt(1), 128)
+	serialNumber, _ := rand.Int(rand.Reader, serialNumberLimit)
+	t := x509.Certificate{
+		SerialNumber: serialNumber,
+		Subject: pkix.Name{
+			Organization: []string{"Acme Co"},
+		},
+		NotBefore: time.Now().Round(time.Second),
+		NotAfter: time.Now().Add(time.Hour).Round(time.Second),
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+	}
+	t.IsCA = true
+	t.KeyUsage |= x509.KeyUsageCertSign
+	der, _ := x509.CreateCertificate(rand.Reader, &t, &t, &rsaTestKey.PublicKey, rsaTestKey)
+	c, _ := x509.ParseCertificate(der)
+	return c
+}()
 
 func RoundtripJWE(keyAlg KeyAlgorithm, encAlg ContentEncryption, compressionAlg CompressionAlgorithm, serializer func(*JsonWebEncryption) (string, error), corrupter func(*JsonWebEncryption) bool, aad []byte, encryptionKey interface{}, decryptionKey interface{}) error {
 	enc, err := NewEncrypter(keyAlg, encAlg, encryptionKey)
--- a/jwk.go
+++ b/jwk.go
@@ -26,7 +26,6 @@ import (
 	"reflect"
 	"strings"
 	"crypto/x509"
-	"github.com/openshift/origin/Godeps/_workspace/src/github.com/coreos/go-oidc/key"
 )
 
 // rawJsonWebKey represents a public or private key in JWK format, used for parsing/serializing.
--- a/jwk.go
+++ b/jwk.go
@@ -21,11 +21,12 @@ import (
 	"crypto/ecdsa"
 	"crypto/elliptic"
 	"crypto/rsa"
+	"crypto/x509"
+	"encoding/base64"
 	"fmt"
 	"math/big"
 	"reflect"
 	"strings"
-	"crypto/x509"
 )
 
 // rawJsonWebKey represents a public or private key in JWK format, used for parsing/serializing.
@@ -56,11 +57,11 @@ type rawJsonWebKey struct {
 
 // JsonWebKey represents a public or private key in JWK format.
 type JsonWebKey struct {
-	Key       interface{}
-	Certificates []x509.Certificate
-	KeyID     string
-	Algorithm string
-	Use       string
+	Key          interface{}
+	Certificates []*x509.Certificate
+	KeyID        string
+	Algorithm    string
+	Use          string
 }
 
 // MarshalJSON serializes the given key to its JSON representation.
@@ -97,7 +98,7 @@ func (k JsonWebKey) MarshalJSON() ([]byte, error) {
 
 	raw.X5c = make([][]byte, len(k.Certificates))
 	for i, cert := range k.Certificates {
-		raw.X5c[i] = []byte(base64URLEncode(cert.Raw))
+		base64.StdEncoding.Encode(cert.Raw, raw.X5c[i])
 	}
 
 	return MarshalJSON(raw)
@@ -134,6 +135,16 @@ func (k *JsonWebKey) UnmarshalJSON(data []byte) (err error) {
 	if err == nil {
 		*k = JsonWebKey{Key: key, KeyID: raw.Kid, Algorithm: raw.Alg, Use: raw.Use}
 	}
+
+	k.Certificates = make([]*x509.Certificate, len(raw.X5c))
+	for i, cert := range raw.X5c {
+		// encoding/json detects and decodes base64 encoding
+		k.Certificates[i], err = x509.ParseCertificate(cert)
+		if err != nil {
+			return err
+		}
+	}
+
 	return
 }
 
--- a/jwk.go
+++ b/jwk.go
@@ -51,12 +51,13 @@ type rawJsonWebKey struct {
 	Dq *byteBuffer `json:"dq,omitempty"`
 	Qi *byteBuffer `json:"qi,omitempty"`
 	// Certificates
-	X5c *byteBuffer `json:"x5c,omitempty"`
+	X5c [][]byte `json:"x5c,omitempty"`
 }
 
 // JsonWebKey represents a public or private key in JWK format.
 type JsonWebKey struct {
 	Key       interface{}
+	Certificates []x509.Certificate
 	KeyID     string
 	Algorithm string
 	Use       string
@@ -76,8 +77,6 @@ func (k JsonWebKey) MarshalJSON() ([]byte, error) {
 		raw, err = fromEcPrivateKey(key)
 	case *rsa.PrivateKey:
 		raw, err = fromRsaPrivateKey(key)
-	case *x509.Certificate:
-		raw, err = fromX509Certificate(key)
 	case []byte:
 		raw, err = fromSymmetricKey(key)
 	default:
@@ -91,6 +90,15 @@ func (k JsonWebKey) MarshalJSON() ([]byte, error) {
 	raw.Kid = k.KeyID
 	raw.Alg = k.Algorithm
 	raw.Use = k.Use
+
+	if len(k.Certificates) == 0 {
+		return MarshalJSON(raw)
+	}
+
+	raw.X5c = make([][]byte, len(k.Certificates))
+	for i, cert := range k.Certificates {
+		raw.X5c[i] = []byte(base64URLEncode(cert.Raw))
+	}
 
 	return MarshalJSON(raw)
 }
@@ -340,21 +348,6 @@ func (key rawJsonWebKey) rsaPrivateKey() (*rsa.PrivateKey, error) {
 
 	err := rv.Validate()
 	return rv, err
-}
-
-func fromX509Certificate(cert *x509.Certificate) (*rawJsonWebKey, error) {
-	var raw *rawJsonWebKey
-	var err error
-
-	switch key := cert.PublicKey.(type) {
-	case *ecdsa.PublicKey:
-		raw, err = fromEcPublicKey(key)
-	case *rsa.PublicKey:
-		raw = fromRsaPublicKey(key)
-	}
-
-	raw.X5c = []byte(base64URLEncode(cert.Raw))
-	return raw, err
 }
 
 func fromRsaPrivateKey(rsa *rsa.PrivateKey) (*rawJsonWebKey, error) {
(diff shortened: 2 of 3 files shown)


jwk.go
```diff
@@ -49,6 +50,8 @@ type rawJsonWebKey struct {
 	Dp *byteBuffer `json:"dp,omitempty"`
 	Dq *byteBuffer `json:"dq,omitempty"`
 	Qi *byteBuffer `json:"qi,omitempty"`
+	// Certificates
+	X5c *byteBuffer `json:"x5c,omitempty"`
```
csstaub: That is incorrect. The `x5c` field is an array of base64-encoded certificates. You can't use `*byteBuffer` here because it encodes to base64url-encoding, which is _not_ the same as base64.

arekkas: ah right, absolutely :) but is the attempt in general ok?

arekkas: btw why is it a byteBuffer and not simply `[]byte`?

jwk.go
```diff
@@ -73,6 +76,8 @@ func (k JsonWebKey) MarshalJSON() ([]byte, error) {
 		raw, err = fromEcPrivateKey(key)
 	case *rsa.PrivateKey:
 		raw, err = fromRsaPrivateKey(key)
+	case *x509.Certificate:
+		raw, err = fromX509Certificate(key)
```
csstaub: I don't think that's what we want -- I don't see the reasoning behind allowing the `Key` field in `JsonWebKey` to support values of type `*x509.Certificate`. This makes it impossible to properly attach a chain of certificates to a key, you would be limited to one. The certificate chain doesn't replace the key, it just supplements it. What you want is to have a `Certificates` field (of type `[]*x509.Certificate`) in `JsonWebKey` which then gets marshaled into `X5c` in the JSON representation of the JWK.

jwk.go
csstaub: Because `[]byte` doesn't marshal according to standard.

jwk.go
arekkas: Ah yeah absolutely, totally missed that.

arekkas:
i hope this makes more sense now :)

csstaub:
Some things:
1) you only added marshal support, no unmarshal support;
2) commit messages should describe the actual changes;
3) your code is not properly gofmt-ed;

and finally you need to sign our [CLA][1] before I can merge changes.

[1]: https://docs.google.com/a/squareup.com/forms/d/13WR8m5uZ2nAkJH41k7GdVBXAAbzDk00vxtEYjd6Imzg/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1

jwk.go
```diff
@@ -87,6 +91,15 @@ func (k JsonWebKey) MarshalJSON() ([]byte, error) {
 	raw.Alg = k.Algorithm
 	raw.Use = k.Use
 
+	if len(k.Certificates) == 0 {
+		return MarshalJSON(raw)
+	}
+
+	raw.X5c = make([][]byte, len(k.Certificates))
+	for i, cert := range k.Certificates {
+		raw.X5c[i] = []byte(base64URLEncode(cert.Raw))
```
csstaub: This is the wrong encoding. Certificates needs to use base64 encoding, which is different from base64url encoding.

csstaub:
Also: this needs tests.

arekkas:
> Some things:
1) you only added marshal support, no unmarshal support;
2) commit messages should describe the actual changes;
3) your code is not properly gofmt-ed;

The point of the commits was to see if the approach is the right one, not to commit a finished version. That's the point of having "unstaged" commit messages and why I pointed out that this is a "draft". Rest will follow soon :) Don't worry, it will be CLA signed and squashed and with meaningful messages. Thanks for pointing me in the right direction.

jwk.go
arekkas: Right, didn't know that, thanks.

arekkas:
ok so both marshalling and unmarshalling have been added. a basic test case in `cookbookJWKs` as well. Does this look legit to you? Should I add a more explicit test?

arekkas:
we could add weak support for `x5t`, `x5t256` and `x5u` by adding fields to `rawJsonWebKey` and `JsonWebKey`. not sure if that was what you intended all along :)

jwk.go
```diff
@@ -121,6 +135,16 @@ func (k *JsonWebKey) UnmarshalJSON(data []byte) (err error) {
 	if err == nil {
 		*k = JsonWebKey{Key: key, KeyID: raw.Kid, Algorithm: raw.Alg, Use: raw.Use}
 	}
+
+	k.Certificates = make([]*x509.Certificate, len(raw.X5c))
+	for i, cert := range raw.X5c {
+		// encoding/json detects and decodes base64 encoding
```
csstaub: Keep in mind this library uses a forked version of `encoding/json` by default.

csstaub:
Don't need to worry about `x5t` et al. for now. Can you add a test for marshaling though? The current tests for `cookbookJWKs` only test unmarshaling.

jwk.go
arekkas: ah right, I was wondering why that worked..

arekkas:
In the latest push, tests fail becuase:

```
jwk_test.go:178: problem unmarshalling asn1: structure error: tags don't match (16 vs {class:1 tag:13 length:73 isCompound:false}) {optional:false explicit:false application:false defaultValue:<nil> tag:<nil> stringType:0 timeType:0 set:false omitEmpty:
false} certificate @2
```

The certificate looks ok though:

```
jwk_test.go:173: JsonBAR: {"kty":"RSA","kid":"bar","alg":"foo","n":"xG3j347FqgBaQHeSTKkuRlOpEVfH6rVy-udv3r7rv25eVaefKxWMDow4C_7TdYTUWZ8JyaQXDYx2vdNyialMASub_N1DBzPvN_i2p4HXtoj_IKOUWZ8yTOVKbwTDRa721jBXGXgoQ2RL3_aekexeX8W_0hTgL1i-xnDvl6y99y69O9umhzfP5TcDY
MRZ1RVf4fmrkI03FgG1JdbRDT0v-o7VJwrF6u4SwNUeF6Du-4ViAjg0SIkWwAenQt-CvEOvVWLbU51MbY_tPvBJQqpQVPMnwWEtJcxC1Xdz5gD2Nm7oqb8knRztCTVfZn99tRbX2YjtrBfRdlEpe9v4N_dIvw","e":"AQAB","d":"XLaViM5lo8D42FbXudFcoxsXwLTTCohMVfrpBZTNxxQ7-Gfeq8JG0hMC3T6Q5nYkmIyFXSRTMTTsUtq7xBghZK
gFv0tcOMkdjyiYZeXWQDTXXrt26PpIDp1nDZGM7ZYQsbYtVjstnel4GqTX4X5UaRX4XRs3WYGkwhtKKygAnesg2wMzOM6VtdWb_uWQPdWnJk2GkQ8em8f0zW0sikuZHJiHFR5Hik-cBDUBBi1stw_JKoleG-ulz2KlQS_to2AFuN0JLhd3t3Ri6mMrUiLyFJu38tetWG0KEWMCLCrv4x5mxv60cilFSIvVbTU8aYbrNIEgR8OOruE9555nPb6UgQ","p"
:"6EEJYXaJ6u1RKZHgU7nCmzDEFr9O46fDLNx5dPTmgF_0ev3w1ZYNH7fhl8-s7xIduL1CyUUVdsP2cqwcpbir43iymjIhy5cFJNsnuMkhk0irWI0rhwMxMyAvjrzd2Bai5tCt_hCZFnNNRhfARRNPYbnaQG3hTtpLoCroJPDzjsE","q":"2IMvCo_5KO9tzSAOfrE8GcLW-uE8NcMeJMxXD-DBlUKMv8rLOqLtjyducgDkYCP2Hx3CZDVQ960k2wLrB
ykzobat1mURF7nQwLyEewM-Zu6fBEoefcXDTaARpmLaTi6kPfxo2v-zuB1y_o8z7owvOCAG9UUAJEsIpmHsgrF4N38","x5c":["TUlJQzVqQ0NBYzZnQXdJQkFnSVFCRm1lMW1vaGl4Q2RPWk5zWkpYRTB6QU5CZ2txaGtpRzl3MEJBUXNGQURBU01SQXdEZ1lEVlFRS0V3ZEJZMjFsSUVOdk1CNFhEVEUyTURZeE1ERXhOVEF3TlZvWERURTJNRFl4T
URFeU5UQXdOVm93RWpFUU1BNEdBMVVFQ2hNSFFXTnRaU0JEYnpDQ0FTSXdEUVlKS29aSWh2Y05BUUVCQlFBRGdnRVBBRENDQVFvQ2dnRUJBTVJ0NDkrT3hhb0FXa0Iza2t5cExrWlRxUkZYeCtxMWN2cm5iOTYrNjc5dVhsV25ueXNWakE2TU9BdiswM1dFMUZtZkNjbWtGdzJNZHIzVGNvbXBUQUVybS96ZFF3Y3o3emY0dHFlQjE3YUkveUNqbEZtZk
1remxTbThFdzBXdTl0WXdWeGw0S0VOa1M5LzJucEhzWGwvRnY5SVU0QzlZdnNadzc1ZXN2ZmN1dlR2YnBvYzN6K1UzQTJERVdkVVZYK0g1cTVDTk54WUJ0U1hXMFEwOUwvcU8xU2NLeGVydUVzRFZIaGVnN3Z1RllnSTRORWlKRnNBSHAwTGZncnhEcjFWaTIxT2RURzJQN1Q3d1NVS3FVRlR6SjhGaExTWE1RdFYzYytZQTlqWnU2S20vSkowYzdRazF
YMlovZmJVVzE5bUk3YXdYMFhaUktYdmIrRGYzU0w4Q0F3RUFBYU00TURZd0RnWURWUjBQQVFIL0JBUURBZ0trTUJNR0ExVWRKUVFNTUFvR0NDc0dBUVVGQndNQk1BOEdBMVVkRXdFQi93UUZNQU1CQWY4d0RRWUpLb1pJaHZjTkFRRUxCUUFEZ2dFQkFCaEtXU09vNEJrSWRtOVF6SDZUSzNjbkNGOGMzQk1YQkRMODNjbDltVVp6L0F0WmxZUSs4NFk1
engyNEtMckRHRGZaNUpjK25remptY1RBdDUyVzMvK0tyb1hyTWk3VGVUK3lMVzlBZXRzdCtxcUF4aEZwaDlDRFRkZVNTQ1kxWGk1cDRKOVlPQXNsNnd3eUdFbk9qaHprbXAyOUlWWGx6RGNYUEg5V25Gdzd2a3VLNzNyeXVmNnpZaG9nbytoTzhWd0JhR1FGWTBrK0VwbkdoMjZUaFJFZWJTWnNLRGRtbzMva0JrKytuZVI0REw1K01qTFN5TXZaaHhBQ
2ZZU2h6bmNOUGw0VlB1SHJPaXh4c1dXK3hvSGdLRDdMb3NScCtkRmttd0xFVlRyT3RGSXMvVTV2dHh4S3RBaEhnQ3dyajFLQTNhdk5ReUYzYjRpSVN4VWpxSzQ9"]}
```

Do you maybe have an idea what the issue is?

crypter_test.go
```diff
@@ -33,6 +37,26 @@ var rsaTestKey, _ = rsa.GenerateKey(rand.Reader, 2048)
...
+		NotBefore: time.Now().Round(time.Second),
+		NotAfter: time.Now().Add(time.Hour).Round(time.Second),
+		KeyUsage:              x509.KeyUsageKeyEncipherment | x509.KeyUsageDigitalSignature,
+		ExtKeyUsage:           []x509.ExtKeyUsage{x509.ExtKeyUsageServerAuth},
+		BasicConstraintsValid: true,
+	}
+	t.IsCA = true
+	t.KeyUsage |= x509.KeyUsageCertSign
+	der, _ := x509.CreateCertificate(rand.Reader, &t, &t, &rsaTestKey.PublicKey, rsaTestKey)
+	c, _ := x509.ParseCertificate(der)
```
arekkas: I'm not sure if that's the right way to create a x509 certificate. I don't understand why `CreateCertificate` doesn't update the values of the template pointer. Maybe I'm doing something wrong

jwk_test.go
```diff
@@ -156,6 +157,40 @@ func TestRoundtripEcPrivate(t *testing.T) {
...
+	if err != nil {
+		t.Error("problem marshaling", err)
+	}
+
+	t.Logf("JsonBAR: %s", jsonbar)
+
+	var jwk2 JsonWebKey
+	err = jwk2.UnmarshalJSON(jsonbar)
+	if err != nil {
+		t.Error("problem unmarshalling", err)
```
arekkas: Tests fail here

crypter_test.go
alokmenghrajani: If you want to learn more about certificate manipulation in Go, take a look at certstrap: https://github.com/square/certstrap/blob/master/pkix/cert_host.go

csstaub: No need to create a certificate on the fly, just throw an hard-coded one in there. 

Here's one we can use for testing:
```
MIIDYTCCAkmgAwIBAgIJAKNePtFE5+2PMA0GCSqGSIb3DQEBCwUAMD4xCzAJBgNV
BAYTAlVTMQswCQYDVQQIEwJDQTETMBEGA1UEChMKZ2hvc3R1bm5lbDENMAsGA1UE
CxMEcm9vdDAeFw0xNjAyMjIxOTE5NTNaFw0xODExMTgxOTE5NTNaMFExCzAJBgNV
BAYTAlVTMQswCQYDVQQIEwJDQTETMBEGA1UEChMKZ2hvc3R1bm5lbDEPMA0GA1UE
CxMGY2xpZW50MQ8wDQYDVQQDEwZjbGllbnQwggEiMA0GCSqGSIb3DQEBAQUAA4IB
DwAwggEKAoIBAQDARiaDcWQc0TRoncxdF8M/l4OPmor4nVf7e1JLOnMaJiwrIvdG
tRMnkM+wJKtIlSg69t2kN3UsuFA6MOhxHyFYxxX69EfV/foMqDHG/+2BwhCd1iQ6
8XoKh5Plu+ItNv1nJaTfBtw2XcaR2m3oiiwbl+6eYNWWDwiNKsj9R+HN8YDPWq4P
UYf9SLycX85ZIqjNv4mtVdLjtFvksH3kD1aUNTO29dTtAyVEPyeYDbkTcDzU6oPK
sklRO4YOJyr5y9BgTVXBRr2GTbeVBF06n94zGrKc1Y54tDLYwXgkcP6MAP9/QiWv
xHPPdbsGB8Y67AQSrydj6qJmq+kA4s/AwYlRAgMBAAGjTzBNMB0GA1UdJQQWMBQG
CCsGAQUFBwMCBggrBgEFBQcDATAsBgNVHREEJTAjhwR/AAABhxAAAAAAAAAAAAAA
AAAAAAABgglsb2NhbGhvc3QwDQYJKoZIhvcNAQELBQADggEBAJkbNP6zGZhOhuLU
+AuNHbs2usiJwyM1nZu8yIXtL/NdxjJCoImSVtiCxBIFoQaGVlyDaRZyr+HtlMEf
a62Go8DrJXd0ZsPto+sByOKQPafbV74vrgVhRUWBTXmVqzfYcEE9XYxRXrJofzMz
7nYl/zGa/pe+9cB68+s9Gi5zvtmP9c0nG2JLr5FakD3lCCfCA6tCDiCGfYcIR+f+
v9K2vTzGRqQu0W+u3eEcEi2CVlIml09vBcRKNSC9j3qQtQDYAfHsAX24kXpPEKDC
s4Rmeph13uR3dhFxbfbeQFIrDvZB4GOI19zV3hqYd8HYqY49WSaOzwGK8hAm+OtJ
o23Dhfs=
```

jwk_test.go
csstaub: Fixed version: #97

arekkas:
superseded in #97

PR #99, opened 2016-06-21
jwt: expose key headers

### LINKED ISSUE #98: jwt: expose key headers
Currently the "jwt" package does not expose the jose.Header fields of the JSONWebToken. This means you can't match on keyIDs to determine which key pair signed the token (and verify appropriately).

I propose adding a "Signatures" field to the JSONWebToken to mirror the JSONWebSignature.

ericchiang:
When creating or parsing JWTs, copy the headers from the encryption
or signature onto the JSONWebToken struct and export them for
package users.

Will add tests to this PR, but wanted to make sure this seemed like a
reasonable approach.

Closes #98


csstaub:
Seems good to me 👍

ericchiang:
Tests added.

csstaub:
Looks great, thank you!

PR #111, opened 2016-09-22
Improve multi-recipient/multi-sig handling

csstaub:
Improvements to multi-recipient/multi-sig handling.

* Remove support for multi-recipient/multi-sig in `Decrypt` and `Verify`.
* Introduce new `VerifyMulti` and `DecryptMulti` functions for old use-cases.
* Improve documentation around multi-recipient/multi-sig a bit.

The new `VerifyMulti` and `DecryptMulti` functions will return the index of the recipient/signature that has been verified, along with the header or signature object. This ensures that callers don't accidentally verify one signature but then read headers from another, as the old `Verify` or `Decrypt` didn't say which recipient/signature ended up being valid. 

Because this is a subtle API change, it will likely be released as `v1.1.0` once merged into the `v1` branch. I'm not aware of any public projects using this library that depend on multi-sig so I think it's acceptable to still call this `v1`.

### DIFF
--- a/crypter.go
+++ b/crypter.go
@@ -19,6 +19,7 @@ package jose
 import (
 	"crypto/ecdsa"
 	"crypto/rsa"
+	"errors"
 	"fmt"
 	"reflect"
 )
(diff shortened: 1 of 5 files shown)


crypter.go
```diff
@@ -323,27 +330,87 @@ func (obj JsonWebEncryption) Decrypt(decryptionKey interface{}) ([]byte, error)
...
+		return -1, JoseHeader{}, nil, ErrCryptoFailure
 	}
 
 	// The "zip" header parameter may only be present in the protected header.
 	if obj.protected.Zip != "" {
 		plaintext, err = decompress(obj.protected.Zip, plaintext)
 	}
 
-	return plaintext, err
+	return index, headers.sanitized(), plaintext, err
```
alokmenghrajani: It might be cleaner to not return index, headers and plaintext if err is set here.

crypter_test.go
```diff
@@ -282,45 +282,46 @@ func TestMultiRecipientJWE(t *testing.T) {
...
-	output, err = parsed.Decrypt(sharedKey)
+	i, _, output, err = parsed.DecryptMulti(sharedKey)
 	if err != nil {
-		t.Error("error on decrypt with AES: ", err)
-		return
+		t.Fatal("error on decrypt with AES: ", err)
+	}
+
+	if i != 1 {
+		t.Fatal("recipient index should be 0 for RSA key")
```
alokmenghrajani: s/0/1/ ?

crypter.go
csstaub: Code here will not execute if decryption failed (will return on the plaintext is nil check above).

PR #117, opened 2016-10-12
Benchmark JWT Encode/Decode

shaxbee:
Sorry for spamming with PRs.

* Added benchmarks for Builder and JSONWebToken - seems like claims marshal/unmarshal roundtrip is not a bottleneck.
* Added examples of validation
* Added tests for parsing RSA signed/encoded tokens
* Added missing copyright header in jwt/examples_test.go
* Added Audience.Contains method
* Audience claim ("aud") is compared ignoring the order of values

### DIFF
--- a/jwt/builder_test.go
+++ b/jwt/builder_test.go
@@ -28,6 +28,7 @@ import (
 	"reflect"
 	"sort"
 	"testing"
+	"time"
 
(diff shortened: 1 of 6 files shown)


csstaub:
Looks good 👍 . Thanks for all your PRs.

Issue #119, opened 2016-10-26
Support Proof-of-Possession Key Semantics for JSON Web Tokens (JWTs) (RFC 7800) cnf claim


zamicol:
Thanks to @shaxbee's additions (#114), there's now JWT claims on the v2 branch.

[RFC 7800](https://tools.ietf.org/html/rfc7800) defines one more claim: cnf.  Now seems like a good time to include it. 

"The value of the "cnf" claim is a JSON object and the members of that object identify the proof-of-possession key."

[Here's the list of JWT claims.](http://www.iana.org/assignments/jwt/jwt.xhtml#IESG)

thalesfsp:
@zamicol @csstaub What's the status of this issue? Thank you!

PR #120, opened 2016-10-26
add confirmation claim to Claims struct

zamicol:
Idea for #119.

### DIFF
--- a/jwt/claims.go
+++ b/jwt/claims.go
@@ -32,6 +32,7 @@ type Claims struct {
 	NotBefore NumericDate `json:"nbf,omitempty"`
 	IssuedAt  NumericDate `json:"iat,omitempty"`
 	ID        string      `json:"jti,omitempty"`
+	Confirm   interface{} `json:"cnf,omitempty"`
 }
 
 // NumericDate represents date and time as the number of seconds since the


csstaub:
Is this all that's needed? No validation or anything? Also `interface{}` is not very useful as a type

csstaub:
Just briefly looked at RFC 7800 it looks like the `cnf` claim can contain `jwk` (Section 3.2) for public keys, `jwe` (Section 3.3) for wrapped symmetric keys, `kid` (Section 3.4) for key ids, `jku` (and `kid`) for a URL representation.

Those values should be represented by corresponding structs in the code, and we should have custom `UnmarshalJSON` implementations for them to parse & sanity check them appropriately. For example, when parsing a JWK from a `jwk` field we should ensure that the JWK represents a public key only (and not a private/symmetric key, for example). For for other values. 

Also, I'm not convinced about the utility of `jku`... that's just asking for trouble. I guess we could expose it as a `url.URL` (from net/url) in case anyone wants it, but let's avoid making any network requests in the library itself.

shaxbee:
I'll be able to work on this after v2 release, probably only jwk field will be used. I wonder if the additional registered claims shouldn't have their own structs, instead of adding stuff to existing jwt.Claims.

zamicol:
Sorry for being aloof, I should be able to get back to you tonight.  

Section 6.2 mentions the confirmation methods registry which can be found here: https://www.iana.org/assignments/jwt/jwt.xhtml#confirmation-methods

So only these four: "jwk", "jwe",  "jku", and "kid".

PR #122, opened 2016-10-27
Update README to links to go-jose.v2

csstaub:
@shaxbee @mcpherrinm @alokmenghrajani @worldwise001 

Updates README to "publish" version 2. Already importable via gopkg.in. Going to add an official v2.0.0 tag later. I removed the examples because they differ between v1/v2, just links to documentation now instead.


csstaub:
Rendered README: https://github.com/square/go-jose/blob/cs/update-readme/README.md

PR #124, opened 2016-11-13
Support non-pointer JSONWebKey in NewSigner

shaxbee:


ericchiang:
Verify as well?

shaxbee:
Added Verify / NewEncrypter / Decrypt

signing.go
```diff
@@ -105,6 +105,8 @@ func newVerifier(verificationKey interface{}) (payloadVerifier, error) {
 		return &symmetricMac{
 			key: verificationKey,
 		}, nil
+	case JSONWebKey:
+		return newVerifier(verificationKey.key)
```
csstaub: Should be `verificationKey.Key` here

csstaub:
LGTM, but looks like you have a small typo which is causing the build to fail.

csstaub:
Looks good -- thank you @shaxbee.

PR #125, opened 2016-11-13
Support creating and parsing nested JWTs

shaxbee:


shaxbee:
While working on this PR I was considering using Fluent API for jwt.Parse{Encrypted,Signed,SignedAndEncrypted} family by making them return token only and store error till call to Claims(...).

Current API:
```go
tok, err := jwt.ParseSigned(raw)
if err != nil {
  panic(err)
}

cl := &jwt.Claims{}
if err := tok.Claims(signingKey, cl); err != nil {
  panic(err)
}
```

Fluent API:
```go
cl := &jwt.Claims{}
jwt.ParseSigned(raw).Claims(signingkKey, cl); err != nil {
  panic(err)
}
```

This change would be breaking though.

crypter.go
```diff
@@ -28,6 +28,7 @@ import (
 type Encrypter interface {
 	Encrypt(plaintext []byte) (*JSONWebEncryption, error)
 	EncryptWithAuthData(plaintext []byte, aad []byte) (*JSONWebEncryption, error)
+	Options() *EncrypterOptions
```
csstaub: Why return a pointer here? It looks like the resulting options struct is read-only.

jwt/jwt.go
```diff
@@ -20,16 +20,27 @@ package jwt
 import (
 	"gopkg.in/square/go-jose.v2"
 	"gopkg.in/square/go-jose.v2/json"
+	"strings"
 )
 
 // JSONWebToken represents a JSON Web Token (as specified in RFC7519).
 type JSONWebToken struct {
+	err     error
```
csstaub: So this means that I could have a `JSONWebToken` struct, but the struct doesn't actually represent a valid token. That seems like it might be confusing. Is there a reason it has to be this way?

shaxbee: err was introduced to allow chaining - namely it's used by NestedJSONWebToken.Decrypt, but I can make it return (*JSONWebToken, error) instead.

crypter.go
shaxbee: Sure, will change it, should I expose options in Signer as well?

jwt/jwt.go
shaxbee: Done

csstaub:
Looks good now, thanks! @shaxbee do you have any additional pull requests you want to submit, or should I tag a new release version after merging this?

shaxbee:
No more PRs at the moment.

csstaub:
Great, I will tag this in 2.0.1 when I get home. And thanks again for your contribution!

PR #131, opened 2016-12-02
WIP: X509 headers support in JWK/JWS

shaxbee:


shaxbee:
Pending additional tests to check if x5t, x5t#256 are generated correctly.

shaxbee:
@ekristen Could you test it out?

ekristen:
@shaxbee apologies for the delay, I do appreciate the work, let me try and get this tested now.

ekristen:
@shaxbee seems to work just fine, it would be nice to be able to control what fields make it into the JWK and the signing key headers. For example, if someone includes the certificates they might just want the x5t or the x5t#256 but not the x5c and visa-versa. 

Thoughts on how it could be implemented to allow the person implementing the go-jose library to choose which fields are present and used?

shaxbee:
@ekristen you can set signature fields in JSONWebKey manually using
X509Thumbprint and X509ThumbprintSHA256 functions without setting certs.
When someone includes certificates already the thumbprints string size is
minimal compared to embedded certificates.

ekristen:
Maybe it is out of scope, but it would be nice to be able to provide the private key, certs, etc and still be able to indicate which fields you want in the JWK and the JWS headers.

shaxbee:
I don't want to make jwk struct too complicated, @csstaub what's your take?

csstaub:
Sorry for being a bit absent, was on vacation for a while -- going to take a look at this soon.

signing.go
```diff
@@ -235,6 +240,10 @@ func (obj JSONWebSignature) Verify(verificationKey interface{}) ([]byte, error)
 		return nil, ErrCryptoFailure
 	}
 
+	if err := verifyCerts(&headers); err != nil {
```
csstaub: We should only verify certificates after we've verified the signature, I don't really trust anything that parses X.509 even if it's written in Go.

signing.go
```diff
@@ -262,6 +271,10 @@ func (obj JSONWebSignature) VerifyMulti(verificationKey interface{}) (int, Signa
 			continue
 		}
 
+		if err := verifyCerts(&headers); err != nil {
```
csstaub: Same here -- move this to verify only after we already verify the sig.

signing.go
```diff
@@ -272,3 +285,31 @@ func (obj JSONWebSignature) VerifyMulti(verificationKey interface{}) (int, Signa
 
 	return -1, Signature{}, nil, ErrCryptoFailure
 }
+
+func verifyCerts(data *rawHeader) error {
+	if len(data.X5c) == 0 {
+		return nil
```
csstaub: I would return an error here, then only call this function if `len(data.X5c) > 0`. Otherwise someone calling this accidentally with an empty chain might not get what they want.

dray92:
@csstaub are we expecting to move forward with this PR?

csstaub:
@dray92 I don't think this can be merged as-is, but feel free to pick up this branch and work on it if you want to get this change merged. Probably needs to expose a function to fetch & verify certificates on the JWK given a CA bundle/trust store similar to the API we have for JWS in the base package.

csstaub:
FYI, I'm trying to get this into v3, and x5u/x5t support at least for JWK (though not for JWS) was merged into master in #242.

jwk.go
```diff
@@ -169,6 +184,18 @@ func (s *JSONWebKeySet) Key(kid string) []JSONWebKey {
...
 
+// X509Thumbprint calculates SHA-1 certificate hash for use with x5t claim.
+func X509Thumbprint(cert *x509.Certificate) string {
+	t := sha1.Sum(cert.Raw)
+	return base64.RawURLEncoding.EncodeToString(t[:])
+}
+
+// X509ThumbprintSHA256 calculates SHA-256 certificate hash for use with x5t#256
+func X509ThumbprintSHA256(cert *x509.Certificate) string {
+	t := sha1.Sum(cert.Raw)
```
mitar, line 195: This should be SHA256.

Issue #130, opened 2016-12-01
Support Base64URL-encoded "n" parameter in JSONWebKey


yun-wang:
The JWA [spec](https://www.rfc-editor.org/rfc/rfc7518.txt) says:

> The "n" (modulus) parameter contains the modulus value for the RSA public key.  It is represented as a Base64urlUInt-encoded value.

However, Base64URL-encoded "n" parameter cannot be unmarshalled into [JSONWebKey](https://github.com/square/go-jose/blob/master/jwk.go#L62) in `go-jose`.

Here is what I'm trying to do:
```go
var d jose.JSONWebKeySet
if err := json.Unmarshal(body, &d); err != nil {
	return err
}
```

When the "n" parameter in the json body is Base64URL encoded instead of Base64 encoded, I get the error `illegal base64 data at input byte 343`. I think this is because `encoding/json` package expects a Base64-encoded string when decoding to [`[]byte`](https://golang.org/pkg/encoding/json/#Marshal).

Not sure if I'm missing something?

This can be solved by changing the type of [`N`](https://github.com/square/go-jose/blob/master/jwk.go#L45) in `rawJSONWebKey` to `string`. If I'm right, let me know, I'll gladly submit a PR.

csstaub:
Do you have an example input? In the struct you linked, [`N`](https://github.com/square/go-jose/blob/v2/jwk.go#L45) is of type `*byteBuffer`, which is a special type that deserializes from [base64url](https://github.com/square/go-jose/blob/v2/encoding.go#L139) already.

jsloyer:
@csstaub below is the result from the jwks endpoint.

```
{
	"keys": [{
		"kty": "RSA",
		"n": "AIbv6q2atH5CgUJauw58enJqDFrpl1h7S0ysbisDwga_omYIGIQey_arOTYt28GkI1cfOXYfenmDIRrBgvvU8WNO8MuDBkbQmlSB1_r026JEF_lH7nMXZt8i3F3Dnaqv0M3D7Jm2bxDLdweI_DQkqsT59WkjZmvex6m9bVWRsASB6kH9R9rIHYZRCqZt5d0rM7nCaZMBtkzuv6o9SUc62FllVWVUzJkLEozUTO6MEH281UruSs88ZMn36voQxC-QAChCa9buHU4wUgeNx3yd5HluOJhk3vXhARbV67W95pBcENvagpEiSBeet1OwFg08UaGYhICciK7f-9oc4RmhdSs=",
[...]

PR #132, opened 2016-12-02
Fix base64url decoding

yun-wang:
Addresses the issue raised in #130.

JSON unmarshalling fails with Base64URL type (URLEncoding with padding). I added an attempt to do `URLEncoding` when `RawURLEncoding` fails.

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -25,6 +25,7 @@ import (
 	"io"
 	"math/big"
 	"regexp"
+	"strings"
 )
 
 var stripWhitespaceRegex = regexp.MustCompile("\\s")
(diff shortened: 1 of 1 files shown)


shaxbee:
@yun-wang RawURLEncoding.DecodeString(strings.TrimRight(src, "=")) should suffice: https://play.golang.org/p/dBBP3Ad0DU

yun-wang:
@shaxbee Much better, thanks!

yun-wang:
Closing the PR per https://github.com/square/go-jose/issues/130#issuecomment-264601404.

PR #136, opened 2017-02-08
Add support for custom protected headers

### LINKED ISSUE #56: Custom headers and additional header support

hlandau:
Fixes #56.

This ended up a more substantial change in terms of the innards of the library than I was expecting, but here we are.

Of note:

  - Go's JSON library (which go-jose uses a modified version of)
    can either unmarshal into a struct, or into a map, but not both.
    Unmarshalling into a struct allows the unmarshalling to be performed
    according to the struct field's types. Unmarshalling into a map
    results in a generic map-based representation of JSON constructs.

    go-jose used an internal struct to unmarshal the protected header,
    which fixed it to using a finite set of protected headers. Since
    this is incompatible with allowing arbitrary custom headers, it was
    changed for a map. json.RawMessage is used to defer parsing of the
    values of this map so that they can be properly unmarshalled into
    their proper types, rather than generic maps.

    As such, this commit creates a fairly substantial change to the
    library.

  - Because of this, the processing of many fields is necessarily
    deferred to a later stage, meaning the opportunity for errors now
    occurs at a slightly later stage. A few internal methods which were
    previously guaranteed now have errors; call sites have been adjusted
    accordingly.

  - Because the Type and ContentType fields previously
    supported by go-jose v2 are simply specific headers, there's no need
    for them when a general facility is available. Thus, these fields have
    been removed. Convenience methods are available to specify them in
    Options structs. Because v2 is not yet stabilized, we can get away
    with making breaking changes. (Moreover Type/ContentType had not
    yet shown up in a v2.x.x tag, meaning that it doesn't show up on
    godoc.org, nor is it available to people who simply go get
    gopkg.in/square/go-jose.v2, so reliance is unlikely.)

### DIFF
--- a/crypter.go
+++ b/crypter.go
@@ -22,6 +22,8 @@ import (
 	"errors"
 	"fmt"
 	"reflect"
+
+	"gopkg.in/square/go-jose.v2/json"
 )
 
 // Encrypter represents an encrypter which produces an encrypted JWE object.
@@ -58,11 +60,10 @@ type keyDecrypter interface {
 type genericEncrypter struct {
 	contentAlg     ContentEncryption
 	compressionAlg CompressionAlgorithm
-	contentType    ContentType
-	typ            ContentType
 	cipher         contentCipher
 	recipients     []recipientKeyInfo
 	keyGenerator   keyGenerator
+	extraHeaders   map[HeaderKey]interface{}
 }
 
 type recipientKeyInfo struct {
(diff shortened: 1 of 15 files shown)


hlandau:
cc @jsha; this will probably be useful to you w.r.t. letsencrypt/pebble.

csstaub:
Thank you for your pull request! I'm at home right now so I can't check if you already signed, but note that Square has an [Individual Contributor License Agreement](https://docs.google.com/forms/d/e/1FAIpQLSeRVQ35-gq2vdSxD1kdh7CJwRdjmUA0EZ9gRXaWYoUeKPZEQQ/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1) that must be signed before I can accept contributions.

crypter.go
```diff
@@ -74,8 +75,32 @@ type recipientKeyInfo struct {
 // EncrypterOptions represents options that can be set on new encrypters.
 type EncrypterOptions struct {
 	Compression CompressionAlgorithm
-	Type        ContentType
-	ContentType ContentType
+
+	// Optional map of additional keys to be inserted into the protected header
+	// of a JWS object. Some specifications which make use of JWS like to insert
+	// additional values here. All values must be JSON-serializable.
+	ExtraHeaders map[HeaderKey]interface{}
```
csstaub: This is an API change, but I think that's ok. I'll make a note to myself to bump the minor version when I tag this.

crypter.go
```diff
@@ -74,8 +75,32 @@ type recipientKeyInfo struct {
...
+}
+
+// WithContentType adds a content type ("cty") header and returns the updated
+// EncrypterOptions.
+func (eo *EncrypterOptions) WithContentType(contentType ContentType) *EncrypterOptions {
+	return eo.WithHeader(HeaderContentType, contentType)
+}
+
+// WithType adds a type ("typ") header and returns the updated EncrypterOptions.
+func (eo *EncrypterOptions) WithType(typ ContentType) *EncrypterOptions {
```
csstaub: 👍 I like this pattern.

crypter.go
csstaub: You mention ContentType/Type wasn't tagged yet, but I think ContentType was, but not Type. I can see ContentType in EncrypterOptions here: https://godoc.org/gopkg.in/square/go-jose.v2. It's fine either way though.

csstaub:
Actually I don't have a lot of comments about this, this looks really great. Thank you for doing the work to make the internals more flexible.

hlandau:
CLA signed.

csstaub:
Looks good to me, thanks!

PR #139, opened 2017-02-23
Fix crash when using symmetric JSONWebKey for JWS signing

b1v1r:
When using a symmetric `JSONWebKey` to sign, there is a crash when assigning the `KeyID` as the `recipient.publicKey` is not yet created at:

https://github.com/square/go-jose/blob/v2/signing.go#L180

This just checks `IsPublic()` and if false, creates the structure with the `KeyID`. 

Example (crashes without patch with nil pointer dereference):
```
package main

import (
	"encoding/base64"
	"encoding/json"
	"fmt"
	"os"
	"strings"

	"gopkg.in/square/go-jose.v2"
)

func exit(msg string, err error) {
	fmt.Fprintf(os.Stderr, "[ERROR] %s: %s\n", err)
	os.Exit(1)
}

func main() {
	key := &jose.JSONWebKey{
		KeyID: "abc123",
		Key:   []byte("super secret"),
	}
	if jtext, err := json.MarshalIndent(key, "", "  "); err == nil {
		fmt.Printf("JWS Key: %s\n", jtext)
	}

	signer, err := jose.NewSigner(jose.SigningKey{Algorithm: jose.HS256, Key: key}, nil)
	if err != nil {
		exit("Could not create signer", err)
	}

	jws, err := signer.Sign([]byte("This is a test"))
	if err != nil {
		exit("Could not sign", err)
	}

	compact, err := jws.CompactSerialize()
	if err != nil {
		exit("Could not compact/serialize", err)
	}
	fmt.Printf("JWS Compact Serialized: %s\n", compact)

	b64data := strings.SplitN(compact, ".", 3)
	protected, err := base64.RawURLEncoding.DecodeString(b64data[0])
	if err != nil {
		exit("Could not base64decode", err)
	}
	fmt.Printf("JWS Protected Header: %s\n", string(protected))
}
```

### DIFF
--- a/signing.go
+++ b/signing.go
@@ -177,7 +177,13 @@ func newJWKSigner(alg SignatureAlgorithm, signingKey JSONWebKey) (recipientSigIn
 	if err != nil {
 		return recipientSigInfo{}, err
 	}
-	recipient.publicKey.KeyID = signingKey.KeyID
+	if signingKey.IsPublic() {
+		recipient.publicKey.KeyID = signingKey.KeyID
+	} else {
+		recipient.publicKey = &JSONWebKey{
+			KeyID: signingKey.KeyID,
+		}
+	}
 	return recipient, nil
 }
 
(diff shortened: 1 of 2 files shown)


csstaub:
Looks good -- thank you for fixing this!

PR #143, opened 2017-03-24
Preserve integers when normalizing JWT claims (fixes #142)

### LINKED ISSUE #142: Builder serializes integers as floats, which subsequently fail to Decode.
Assuming I'm doing this all using the blessed path. I've included a snippet at the bottom to demonstrate.

Relevant output produced:
```
serialized to eyJhbGciOiJIUzUxMiJ9.eyJudW1iZXIiOjFlKzA2fQ.oFRp6CQOyNEicaiXDZASuGZpAX00fWnCoMRd89ueIFqo2t48WUo4C3Ld0EkrSViLYk5FxoxmyxXRJaq6BYKWuA
parsing signed
parsing claims
json: cannot unmarshal number 1e+06 into Go value of type int64
```

If you base64 decode the 2nd component of the compact JWT you'll see `"number": 1e+06`.

I suspect the culprit is here: https://github.com/square/go-jose/blob/v2/jwt/builder.go#L132 - the Marshal/Unmarshal loses the type information that would otherwise cause go's marshaler to encode `Number` as an integer.

```
package main

import (
	"fmt"

	jose "gopkg.in/square/go-jose.v2"
	jwt "gopkg.in/square/go-jose.v2/jwt"
)

const SecretKey = "WoW"

type Claims struct {
	Number int64 `json:"number"`
}

func main() {
	fmt.Println("building signer")
	signer, err := jose.NewSigner(jose.SigningKey{Algorithm: jose.HS512, Key: []byte(SecretKey)}, nil)
	if err != nil {
		fmt.Println(err)
		return
	}

	fmt.Println("building builder")
	token := jwt.Signed(signer)
	token = token.Claims(Claims{Number: 1000000})

	fmt.Println("serializing")
	enc, err := token.CompactSerialize()
	if err != nil {
		fmt.Println(err)
		return
	}
	fmt.Printf("serialized to %s\n", enc)

	fmt.Println("parsing signed")
	tok, err := jwt.ParseSigned(enc)
	if err != nil {
		fmt.Println(err)
		return
	}

	fmt.Println("parsing claims")
	claims := Claims{}
	if err := tok.Claims([]byte(SecretKey), &claims); err != nil {
		fmt.Println(err)
		return
	}
	fmt.Printf("%#v\n", claims)
}
```

csstaub:
We normalize claims for tokens given to us in the builder by marshaling and then unmarshaling as a JSON object. However, because JSON officially only supports floating-point numbers, this loses precision for integers embedded in structs (see #142). We switch to using json.Decoder (with UseNumber option) to decode numbers into json.Number, which preserves the precise value for both integers and floats.

Go Playground example to play around with old/new code: https://play.golang.org/p/2irRfWlxgN

cc @shaxbee @jacobstr

### DIFF
--- a/jwt/example_test.go
+++ b/jwt/example_test.go
@@ -25,6 +25,7 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+
 	"gopkg.in/square/go-jose.v2"
 	"gopkg.in/square/go-jose.v2/jwt"
 )
@@ -149,7 +150,7 @@ func ExampleSigned() {
 	}
 
 	fmt.Println(raw)
-	// Output: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOlsibGVlbGEiLCJmcnkiXSwiaXNzIjoiaXNzdWVyIiwibmJmIjoxLjQ1MTYwNjRlKzA5LCJzdWIiOiJzdWJqZWN0In0.mI6U-xUdttpOPIDUAI2uyg9lFgoqaAb-hwmz8L6L3fo
+	// Output: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOlsibGVlbGEiLCJmcnkiXSwiaXNzIjoiaXNzdWVyIiwibmJmIjoxNDUxNjA2NDAwLCJzdWIiOiJzdWJqZWN0In0.4PgCj0VO-uG_cb1mNA38NjJyp0N-NdGIDLoYelEkciw
 }
 
 func ExampleEncrypted() {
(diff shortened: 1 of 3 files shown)


jwt/example_test.go
```diff
@@ -149,7 +150,7 @@ func ExampleSigned() {
 	}
 
 	fmt.Println(raw)
-	// Output: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJhdWQiOlsibGVlbGEiLCJmcnkiXSwiaXNzIjoiaXNzdWVyIiwibmJmIjoxLjQ1MTYwNjRlKzA5LCJzdWIiOiJzdWJqZWN0In0.mI6U-xUdttpOPIDUAI2uyg9lFgoqaAb-hwmz8L6L3fo
```
jacobstr: Cool:
```
{
  "aud": [
    "leela",
    "fry"
  ],
  "iss": "issuer",
  "nbf": 1451606400,
  "sub": "subject"
}```

Issue #68, opened 2016-02-13
CFRG ECDH and signatures in JOSE 


csstaub:
Support for: https://datatracker.ietf.org/doc/draft-ietf-jose-cfrg-curves/

darkk:
As far as I see, support added in #154 is a bit incomplete, `IsPublic()` and `IsValid()` always return `false`.

csstaub:
Sounds like a bug. @darkk could you open a new issue with some steps to reproduce?

darkk:
Heh, seems, I was late. I see that it's already fixed. Thanks, @csstaub! :-)

csstaub:
Yes, this should work now (for ed25519). It's still fairly new, let us know if you run into issues with it.

PR #154, opened 2017-07-24
Add  support for ed25519 signing

rahulredd:
This PR adds support for ed25519 signing and verification. 

Address this issue -  #68

### DIFF
--- a/asymmetric.go
+++ b/asymmetric.go
@@ -28,6 +28,7 @@ import (
 	"fmt"
 	"math/big"
 
+	"golang.org/x/crypto/ed25519"
 	"gopkg.in/square/go-jose.v2/cipher"
 	"gopkg.in/square/go-jose.v2/json"
 )
@@ -47,6 +48,10 @@ type ecEncrypterVerifier struct {
 	publicKey *ecdsa.PublicKey
 }
 
+type edEncrypterVerifier struct {
+	publicKey ed25519.PublicKey
+}
+
 // A key generator for ECDH-ES
 type ecKeyGenerator struct {
 	size      int
@@ -57,6 +62,10 @@ type ecKeyGenerator struct {
 // A generic EC-based decrypter/signer
 type ecDecrypterSigner struct {
 	privateKey *ecdsa.PrivateKey
+}
+
+type edDecrypterSigner struct {
+	privateKey ed25519.PrivateKey
 }
 
 // newRSARecipient creates recipientKeyInfo based on the given key.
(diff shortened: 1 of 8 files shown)


jwk.go
```diff
@@ -379,6 +444,14 @@ func (key rawJSONWebKey) rsaPrivateKey() (*rsa.PrivateKey, error) {
 	return rv, err
 }
 
+func fromEdPrivateKey(ed ed25519.PrivateKey) (*rawJSONWebKey, error) {
+
```
csstaub: Remove extra whitespace here

shared.go
```diff
@@ -68,6 +68,7 @@ var (
 
 // Key management algorithms
 const (
+	ED25519Key         = KeyAlgorithm("ED25519")
```
csstaub: Why not just call this constant `ED25519`?

asymmetric.go
```diff
@@ -438,6 +466,33 @@ func (ctx ecDecrypterSigner) decryptKey(headers rawHeader, recipient *recipientI
 
 	return josecipher.KeyUnwrap(block, recipient.encryptedKey)
 }
+func (ctx edDecrypterSigner) signPayload(payload []byte, alg SignatureAlgorithm) (Signature, error) {
+
```
csstaub: Remove extra whitespace

asymmetric_test.go
```diff
@@ -142,6 +142,45 @@ func TestVectorsRSA(t *testing.T) {
 	}
 }
 
+func TestEd25519(t *testing.T) {
+
```
csstaub: Remove extra whitespace

jwk.go
```diff
@@ -126,6 +132,16 @@ func (k *JSONWebKey) UnmarshalJSON(data []byte) (err error) {
 		}
 	case "oct":
 		key, err = raw.symmetricKey()
+	case "OKP":
+		if raw.Crv == "Ed25519" {
```
csstaub: Could add a check here that X is present, as it must always be present for both private and public keys:
```
if raw.Crv == "Ed25519 && raw.X != "" {
        // parse 
} else {
        // return error
}
```

csstaub:
Hi! Thanks for your contribution! This looks great, I left some minor comments. One thing I think we should add is more test vectors (ideally from the appendix of the spec) to make sure the code generates the correct output given a fixed set of parameters.

rahulredd:
Fixed all the indentation issues and added test vectors.

csstaub:
This looks good. Thank you for your contribution @rahulredd!

csstaub:
If you want me to cut a new release with this, let me know.

PR #159, opened 2017-09-08
Prevent panics when sanitizing header values that may be nil

rolandshoemaker:
`rawHeader.sanitized` doesn't check if the values of the map it is `range`ing over are `nil` before dereferencing them. Since `encoding/json` is happy to unmarshal the value `null` into a `nil` when the field value is a pointer this allows a message to be crafted that will cause a panic, i.e. if the protected header contains `{"alg":"ES256","nonce":null}` it will be unmarshaled into the map as `"alg": ..., "nonce": nil`.

This adds a simple check that the value of the field is non-`nil` before attempting to parse it further and a test that this fixes the panics.

### DIFF
--- a/jws_test.go
+++ b/jws_test.go
@@ -310,3 +310,25 @@ func TestErrorMissingPayloadJWS(t *testing.T) {
 		t.Errorf("unexpected error message, should contain 'missing payload': %s", err)
 	}
 }
+
+// Test that a null value in the header doesn't panic
+func TestNullHeaderValue(t *testing.T) {
+	msg := `{
+   "payload":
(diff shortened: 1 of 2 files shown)


csstaub:
Looks good to me. Thank you for fixing @rolandshoemaker!

cpu:
@csstaub Is there any chance you could cut a release with this fix in the next week? It would be great to get this shipped.

csstaub:
@cpu Definitely, I'll make sure to cut a release soon.

cpu:
@csstaub Great! Thank you :-)

Issue #155, opened 2017-08-04
Improve handling of JWS protected headers vs unprotected headers


jsha:
Right now, protected and unprotected headers get merged into a single headers structure. This is causing an issue for us on https://github.com/letsencrypt/boulder/ as we implement the ACMEv2 API, because v2 uses protected headers exclusively, and we'd like to enforce that none of our fields come from an unprotected header.

This may be indicative of a broader issue: Consumers of JWS in general will accept a header field from either the protected or unprotected header. If the producer (signer) of that JWS is configured to put sensitive fields in the unprotected header, the consumer won't reject it, and everything will appear to work fine, until an attacker modifies the unprotected header and it goes undetected.

Possible solution: Remove merging of headers. Put unprotected headers in a separate `Unprotected` field of `Signature`.

csstaub:
Agree that splitting these is a good idea. We can expose protected/unprotected/merged in separate fields, and consumers can pick which one they'd like to read. Actually maybe we should get rid of the merged header set entirely but that'd break backwards compat, but could be a future change.

csstaub:
Closing this out, was fixed in #164.

PR #164, opened 2017-09-20
Fix issue #155: expose protected/unprotected headers separately

csstaub:
Expose protected/unprotected headers separately so callers can distinguish between values that were protected vs. unprotected. cc @jsha


jsha:
Sorry I forgot to review this earlier, but LGTM. Thanks for implementing!

PR #165, opened 2017-09-29
Use consistent types for ed25519.PrivateKey and ed25519.PublicKey

pquerna:
Since these are already bytes slices, we shouldn't be using pointers to them.  This is different than most of the other key types, since they are generally realized structs.


csstaub:
Looks good to me. Thank you for your contribution @pquerna!

PR #166, opened 2017-10-04
JWK key generator tool

darkk:
I'm adding this key generator as JWK has canonical representation, so util to generate alike keys sounds like something useful to go-jose project.

I'm going to use this key format for my patch for [heipei/nginx-sso](https://github.com/heipei/nginx-sso) so I can also move it there, but I feel like go-jose is right place for this generator.

`Public()` method is added to be able to load private key during application initialization and use it's public part in part of the application that do not require private key, e.g. when the daemon issues signed token to the browser and checks signature in some of code paths.

Also, I've electronically signed [Individual Contributor License Agreement][1].

 [1]: https://spreadsheets.google.com/spreadsheet/viewform?formkey=dDViT2xzUHAwRkI3X3k5Z0lQM091OGc6MQ&ndplr=1

### DIFF
--- a/jwk-keygen/README.md
+++ b/jwk-keygen/README.md
@@ -1,0 +1,35 @@
+# JWK Key Generator
+
+The `jwk-keygen` command line utility generates keypairs used for asymmetric
+encryption and signing algorithms in JSON Web Key (JWK) format.
+
+## Usage
+
+The utility requires specification of both desired algorithm (`alg`) and key
+usage (`use`) to remind that same keypair should never be used both for
+encryption and signing.
+
+Algorithms are selected via the `--alg` flag, which influence the `alg` header.
+For JWE (`--use=enc`), `--alg` specifies the key management algorithm (e.g.
+`RSA-OAEP`).  . For JWS (`--use=sig`), `--alg` specifies the signature
+algorithm (e.g. `PS256`).
+
+Output file is determined by specified usage, algorithm and Key ID, e.g.
+`jwk-keygen --use=sig --alg=RS512 --kid=test` produces files
+`jwk_sig_RS512_test` and `jwk_sig_RS512_test.pub`. Keys are sent to stdout when
+no Key ID is specified: neither pre-defined nor random one.
+
+## Examples
+
+### RSA 2048
+
+Generate RSA/2048 key for encryption and output to stdout.
+
+    jwk-keygen --use enc --alg RSA-OAEP
+
+### Custom key length
+
+Generate RSA/4096 key for signing and store to files.
+
+    jwk-keygen --use sig --alg RS256 --bits 4096 --kid test
+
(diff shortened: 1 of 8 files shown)


csstaub:
Hi @darkk, this is great, thank you! I can review the code in a bit, but before I can merge anything Square requires a CLA to be signed. If you previously signed our CLA through our old Google Form this is the same language, we just recently switched this repo to CLA-assistant.io to make managing it easier. Let me know if you have questions.

darkk:
@csstaub thanks for mentioning that, `CONTRIBUTING.md` still points to google forms and `CLA-assistant.io` was broken for me right after the PR was submitted (it was redirecting back to PR page), so I assumed that it actually checks some table field through google form :-)

csstaub:
Aha, I should probably update the `CONTRIBUTING.md` file. Thanks for pointing that out.

jwk.go
```diff
@@ -254,6 +254,25 @@ func (k *JSONWebKey) IsPublic() bool {
...
+	ret := *k
+	switch key := k.Key.(type) {
+	case *ecdsa.PrivateKey:
+		ret.Key = key.Public()
+	case *rsa.PrivateKey:
+		ret.Key = key.Public()
+	case ed25519.PrivateKey:
+		ret.Key = key.Public()
+	default:
+		return JSONWebKey{} // returning invalid key
```
csstaub: Would prefer if this returned an error, otherwise potentially confusing to callers.

jwk-keygen/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+## Usage
+
+The utility requires specification of both desired algorithm (`alg`) and key
+usage (`use`) to remind that same keypair should never be used both for
+encryption and signing.
+
+Algorithms are selected via the `--alg` flag, which influence the `alg` header.
+For JWE (`--use=enc`), `--alg` specifies the key management algorithm (e.g.
+`RSA-OAEP`).  . For JWS (`--use=sig`), `--alg` specifies the signature
```
csstaub: Small nit, there's a typo here (double period)

jwk-keygen/main.go
```diff
@@ -0,0 +1,183 @@
...
+	if *kid == "" {
+		fmt.Printf("==> jwk_%s.pub <==\n", *alg)
+		fmt.Println(string(pubJS))
+		fmt.Printf("==> jwk_%s <==\n", *alg)
+		fmt.Println(string(privJS))
+	} else {
+		// JWK Thumbprint (RFC7638) is not used for key id because of
+		// lack of canonical representation.
+		fname := fmt.Sprintf("jwk_%s_%s_%s", *use, *alg, *kid)
+		err = ioutil.WriteFile(fname+".pub", pubJS, 0444)
```
csstaub: Should add checks here that file doesn't already exist, otherwise I think `ioutil.WriteFile` will truncate & overwrite anything that already exists at this location. That's potentially dangerous.

csstaub:
Looks good to me, I left some minor comments but nothing major.

jwk.go
darkk: Should it return `(JSONWebKey, error)`, (*JSONWebKey, error)` or `*JSONWebKey` with nil on error? I'm rather new to Go, so I'm unsure what's the most "idiomatic" way.

csstaub:
Looks good now, thank you @darkk!

PR #168, opened 2017-10-27
Supply custom ECDH key and apu/apv claims

NeilMadden:
NB: this PR needs a bit of work to get it into a mergeable state but I wanted to put it up for your feedback. I have the go-ahead from my employer wrt to the contributor agreement.

This adds two new encryptor options: one provides a callback for generating the apu and apv values and is handed the public keys for both parties; the other allows the caller to specify a custom ephemeral key generator.

For an illustration of the kinds of issues that populating the apu and apv claims is meant to solve, see for instance the rationale for the [TLS extended master secret in response to the Triple Handshake vulnerability](https://www.imperialviolet.org/2014/03/03/triplehandshake.html). In this case, including too little information in the key derivation hash process was identified as the root cause and so the extended master secret was developed that hashes in the full transcript of the TLS handshake up to that point. In particular, you should always at least include the public keys in the KDF hash, hence why we make them available to the callback.

The second change to allow a custom ephemeral key to be provided is to support including a signed hash of the ephemeral public key in the payload of the JWE to authenticate the ephemeral public key. Otherwise, the ephemeral key is generated too late in the process for this to happen so we generate it up front and then supply it to the ECDH key agreement process via this callback.

Let me know if you have comments/feedback. I'm not precious about the approach, so happy to rework if you'd prefer a different design. In the meantime, I'll change the imports back and add some tests.

### DIFF
--- a/asymmetric.go
+++ b/asymmetric.go
@@ -31,6 +31,7 @@ import (
 	"golang.org/x/crypto/ed25519"
 	"gopkg.in/square/go-jose.v2/cipher"
 	"gopkg.in/square/go-jose.v2/json"
+	"encoding/base64"
 )
 
 // A generic RSA-based encrypter/verifier
@@ -54,9 +55,10 @@ type edEncrypterVerifier struct {
 
 // A key generator for ECDH-ES
 type ecKeyGenerator struct {
-	size      int
-	algID     string
-	publicKey *ecdsa.PublicKey
+	size               int
+	algID              string
+	publicKey          *ecdsa.PublicKey
+	partyInfoGenerator PartyInfoGenerator
 }
 
 // A generic EC-based decrypter/signer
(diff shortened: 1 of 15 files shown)


NeilMadden:
Apologies - just re-reading the email I got and we are still waiting for a final go-ahead from legal wrt to the CLA. Will push to get a decision on this asap.

csstaub:
Hi @NeilMadden, sorry for the delayed response but I was on vacation for the last two weeks. I will make sure to review your code tomorrow, but in general this seems like the right approach. Let me know if you have questions around the CLA and/or if you want to setup a corp CLA instead of individual CLAs.

crypter.go
```diff
@@ -23,7 +23,8 @@ import (
 	"fmt"
 	"reflect"
 
-	"gopkg.in/square/go-jose.v2/json"
+	"gopkg.in/NeilMadden/go-jose.v2/json"
```
csstaub: Import path needs fixing here.

asymmetric.go
```diff
@@ -29,8 +29,9 @@ import (
 	"math/big"
 
 	"golang.org/x/crypto/ed25519"
-	"gopkg.in/square/go-jose.v2/cipher"
-	"gopkg.in/square/go-jose.v2/json"
+	"gopkg.in/NeilMadden/go-jose.v2/cipher"
+	"gopkg.in/NeilMadden/go-jose.v2/json"
```
csstaub: Import paths need to point to square/go-jose.v2 again

csstaub:
Code looks good to me, apart from the import paths.

csstaub:
Ping @NeilMadden -- any luck on the CLA, wrt. your legal dept? If so I'd be happy to clean this up and merge it.

PR #170, opened 2018-02-09
Allow empty signature

earlonrails:
When I was using this library for work I was getting the error `square/go-jose: error in cryptographic primitive`. Firstly we use the [this](https://github.com/jwt/ruby-jwe) ruby library on our rails server. We have an asymmetric key pair and encrypt JWTs like something like this 
```
# no password on the jwt
jwt = JWT.encode(self, nil, 'none')
# => "eyJ0eXAiOiJKV1QiLCJhbGciOiJub25lIn0.eyJpc3MiOiJPQXV0aCIsImV4cCI6MTUxODE1MTQwM30."
jwe = JWE.encrypt(jwt, public_key)

```
Even on their [readme](https://github.com/jwt/ruby-jwt#algorithms-and-usage) it shows a similar jwt,
`eyJ0eXAiOiJKV1QiLCJhbGciOiJub25lIn0.eyJkYXRhIjoidGVzdCJ9.`

But when I would try to use this library I was getting an error.

```
 var noPassword []byte
  tok, err := jose.ParseEncrypted(s)
  if err != nil {
    fmt.Printf("error parsing jwe: %v ", err)
  }

  nested, err := tok.Decrypt(d.PrivateKey)
  if err != nil {
    fmt.Printf("error decrypting jwe: %v ", err)
  }

  wt, err := jwt.ParseSigned(string(nested))
  if err != nil {
    fmt.Printf("error serializing into jwt: %v ", err)
    return user, err
  }

  claims := make(map[string]interface{})
  if err := wt.Claims(noPassword, &claims); err != nil {
    fmt.Printf("error serializing claims: %v ", err)
    // Here I was getting the error square/go-jose: error in cryptographic primitive
  }
```
I might be doing things wrong here and if so let me know, but this fixes things for my use case.

### DIFF
--- a/jwt/claims_test.go
+++ b/jwt/claims_test.go
@@ -78,3 +78,16 @@ func TestDecodeClaims(t *testing.T) {
 		assert.Equal(t, v.Err, json.Unmarshal([]byte(v.Raw), &c))
 	}
 }
+
+func TestEmptySignature(t *testing.T) {
+	var empty []byte
+	s := "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiIxMjM0NTY3ODkwIiwibmFtZSI6IkpvaG4gRG9lIn0."
+	wt, err := ParseSigned(s)
+	if err != nil {
+		t.Errorf("Failed to parse jws: %#v", err)
+	}
+	out := make(map[string]interface{})
+	if err := wt.Claims(empty, &out); err != nil {
+		t.Errorf("Claim with no signature failed: %#v", err)
+	}
+}
--- a/signing.go
+++ b/signing.go
@@ -302,6 +302,9 @@ func (obj JSONWebSignature) Verify(verificationKey interface{}) ([]byte, error) 
 
 	input := obj.computeAuthData(&signature)
 	alg := headers.getSignatureAlgorithm()
+	if len(signature.Signature) == 0 {
+		return obj.payload, nil
+	}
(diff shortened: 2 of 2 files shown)


csstaub:
Allowing tokens without signatures to be accepted compromises the security model. An attacker could circumvent any signature on a token by just stripping it. It's true that in your case the token happens to be encrypted, but that's not true in the general case. I also think the 'none' algorithm for JWTs was a misfeature of the standard and should not be supported. If you don't need the integrity protection there is no reason to use a JWT, you can just put your payload directly inside the JWE.

earlonrails:
I was looking at the rfc for jwt and I found this.
https://tools.ietf.org/html/rfc7523
```
  9.   The JWT MUST be digitally signed or have a Message
        Authentication Code (MAC) applied by the issuer.  The
        authorization server MUST reject JWTs with an invalid signature
        or MAC.
```
Therefore, if someone is using a MAC then this would totally be valid to not have a signature.
In our case we use a JTI, but others could use something else in the message.
https://en.wikipedia.org/wiki/Message_authentication_code
Thanks for your consideration!

csstaub:
In the case of a MAC-ed message the `signature` will contain the computed MAC value. An empty (or missing) `signature` means that the JWS was unsecured. This is only ever the case if `alg` was set to `none`, which I do not intend to support. See Appendix A.1 and and A.5 of RFC 7515 for examples.

earlonrails:
Ah ok, just thought I would give it one more try 🏄  Thanks for explanation!

PR #178, opened 2018-03-28
Support certificate chains in headers

csstaub:
Add support for parsing certificate chains from `x5c` headers in messages, and expose a function to verify & read them. The API here is that we expose a new function `jose.Header#Certificates(opts x509.VerifyOptions)` that both reads, verifies & returns the certificate chain a header if present. This forces callers to verify the certificates, which I think is a nice property.

### DIFF
--- a/shared.go
+++ b/shared.go
@@ -160,15 +160,37 @@ type rawHeader map[HeaderKey]*json.RawMessage
 
 // Header represents the read-only JOSE header for JWE/JWS objects.
 type Header struct {
-	KeyID        string
-	JSONWebKey   *JSONWebKey
-	Algorithm    string
-	Nonce        string
-	Certificates []*x509.Certificate
+	KeyID      string
+	JSONWebKey *JSONWebKey
+	Algorithm  string
+	Nonce      string
 
-	// Any headers not recognised above get unmarshaled from JSON in a generic
-	// manner and placed in this map.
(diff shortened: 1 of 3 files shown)


shared.go
```diff
@@ -162,11 +165,34 @@ type Header struct {
...
+// not be validated with the given verify options.
+func (h Header) Certificates(opts x509.VerifyOptions) ([][]*x509.Certificate, error) {
+	if len(h.certificates) == 0 {
+		return nil, errors.New("square/go-jose: no x5c header present in message")
+	}
+
+	leaf := h.certificates[0]
+	if opts.Intermediates == nil {
+		opts.Intermediates = x509.NewCertPool()
+		for _, intermediate := range h.certificates[1:] {
```
mpuncel: is there definitely >=1 intermediate? otherwise this might panic at runtime

csstaub: Excellent point I should fix that, there could definitely be zero here

csstaub: Oh, actually it's fine: there's at least one element and `c[1:]` will just be an empty slice:
```
❯❯❯ gore
gore version 0.2.6  :help for help
gore> c := []string{"test"}
[]string{
  "test",
}
gore> c[1:]
[]string{}
```

csstaub: Added some extra test cases to cover this

PR #179, opened 2018-03-30
support asymmetric signing and verification with opaque keys

mikedanese:
This enables signing with crypto oracles.

### DIFF
--- a/cryptosigner/cryptosigner.go
+++ b/cryptosigner/cryptosigner.go
@@ -1,0 +1,138 @@
+/*-
+ * Copyright 2018 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// Package cryptosigner implements an OpaqueSigner that wraps a "crypto".Signer
+//
+// https://godoc.org/crypto#Signer
+package cryptosigner
+
+import (
+	"crypto"
+	"crypto/ecdsa"
+	"crypto/rand"
+	"crypto/rsa"
+	"encoding/asn1"
+	"io"
+	"math/big"
+
+	"golang.org/x/crypto/ed25519"
+	"gopkg.in/square/go-jose.v2"
+)
+
+// Opaque creates an OpaqueSigner from a "crypto".Signer
+func Opaque(s crypto.Signer) jose.OpaqueSigner {
+	pk := &jose.JSONWebKey{
+		Key: s.Public(),
+	}
+	return &cryptoSigner{signer: s, rand: rand.Reader, pk: pk}
+}
+
+type cryptoSigner struct {
+	pk     *jose.JSONWebKey
+	signer crypto.Signer
+	rand   io.Reader
+}
+
+func (s *cryptoSigner) Public() *jose.JSONWebKey {
+	return s.pk
+}
+
+func (s *cryptoSigner) Algs() []jose.SignatureAlgorithm {
+	switch s.signer.Public().(type) {
+	case ed25519.PublicKey:
+		return []jose.SignatureAlgorithm{jose.EdDSA}
+	case *ecdsa.PublicKey:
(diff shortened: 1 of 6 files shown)


csstaub:
This looks great, thanks @mikedanese!

csstaub:
I'll try to get this merged later today and probably update the README and/or add some examples in the doc for this. Do you want me to tag a release with your changes once it's merged?

mikedanese:
Thanks. I'm in no rush to get this in. I'm going to propose an alternative API today that I think might be cleaner.

mikedanese:
I prefer the API introduced in the 1st commit. It's more flexible. The API I proposed yesterday is here for reference:

https://github.com/square/go-jose/compare/v2...mikedanese:oracle-signer-old?expand=1

I moved the crypto.Signer to an example. I don't think it needs to be exported by this library since now it can be implemented elsewhere.

csstaub:
How about exposing the OpaqueSigner implementation for crypto.Signer in a sub-package? Seems useful to have it available rather than having folks re-implement themselves.

mikedanese:
To test this properly, we'll either have to copy some of the roundtrip testing code in the main package or move it to a testing package. I went with the copy to start with.

mikedanese:
Looked at the tests for cryptosigner package again. The round trip testing is pretty entangled with jose package. I recommend we go with copying. The other options are:
* move this package functionality into the jose package
* do a large refactor of jose package testing including touching randReader to break cylic dependencies.

csstaub:
I'm ok with copying, that sounds fine to me. And thank you for your work on this! Let me know when you think this is ready for merging/you want me to review more closely.

cryptosigner/cryptosigner.go
```diff
@@ -0,0 +1,146 @@
...
+	"encoding/base64"
+	"io"
+	"math/big"
+
+	"golang.org/x/crypto/ed25519"
+	"gopkg.in/square/go-jose.v2"
+)
+
+// OpaqueSignerFromCryptoSigner creates a OpaqueSigner from a "crypto".Signer
+func OpaqueSignerFromCryptoSigner(s crypto.Signer) jose.OpaqueSigner {
```
cjcullen: This function name is a little unwieldy (especially in the `cryptosigner` pkg, where it takes a `crypto.Signer`)

Maybe just `Opaque(s crypto.Signer) jose.OpaqueSigner`?

cryptosigner/cryptosigner.go
```diff
@@ -0,0 +1,146 @@
...
+	pk := &jose.JSONWebKey{
+		Key: s.Public(),
+	}
+	b, err := pk.MarshalJSON()
+	if err != nil {
+		panic(err)
```
cjcullen: Is `panic()` the best we can do? Would there be any value in letting a caller know "hey, your signer had an unmarshallable key"?

cryptosigner/cryptosigner.go
```diff
@@ -0,0 +1,146 @@
...
+	case jose.RS512, jose.PS512, jose.ES512:
+		hash = crypto.SHA512
+	default:
+		return nil, jose.ErrUnsupportedAlgorithm
+	}
+
+	var hashed []byte
+	if hash != crypto.Hash(0) {
+		hasher := hash.New()
+		// According to documentation, Write() on hash never fails
```
cjcullen: I believe in the intent of the interface, and I trust the implementations your using, but no interface can guarantee to me that an implementation will never produce an error :).

cryptosigner/cryptosigner.go
```diff
@@ -0,0 +1,146 @@
...
+		var b []byte
+		b, err = s.signer.Sign(s.rand, hashed, hash)
+		if err != nil {
+			return nil, err
+		}
+
+		sig := struct {
+			R, S *big.Int
+		}{}
+		_, err = asn1.Unmarshal(b, &sig)
```
cjcullen:
```
if _, err := ... ; err != nil {
```

opaque.go
```diff
@@ -0,0 +1,83 @@
...
+// OpaqueSigner is an interface that supports signing payloads with opaque
+// private key(s). Private key operations preformed by implementors may, for
+// example, occur in a hardware module. An OpaqueSigner may rotate signing keys
+// transparently to the user of this interface.
+type OpaqueSigner interface {
+	// Public returns the public key of the current signing key.
+	Public() *JSONWebKey
+	// Algs returns a list of supported signing algorithms.
+	Algs() []SignatureAlgorithm
+	// SingPayload signs a payload with the current signing key using the given
```
cjcullen: SignPayload

mikedanese:
@csstaub please take a look

cjcullen:
LGTM

csstaub:
Looks good to me, thanks @mikedanese.

Issue #169, opened 2018-01-24
No way to read issuer without verifying a JWT


mikedanese:
I would like to read the issuer claim of a JWT before and possibly without verifying the signature of the JWT. For this I need some access to an unverified payload. I understand that the library is designed to disallow access to unverified JWS content, but it prevents some reasonable usecases. Is this not possible with the current API or am I missing something? Is it possible to support this usecase while still minimizing the potential for programmer error/security issues?

pquerna:
FWIW we also have a use case where we need to access the unverified headers.  Adding accessors to `JSONWebSignature` would be minimally helpful;  

We can do this in our own code of course: Hacking together a compact form parser isn't that hard -- just period splitting / base64 decoding, but would prefer to keep things in the jose.v2 types if I could.

PR #181, opened 2018-05-13
A potential fix for #169: Get claims from jwt even if validation fails

bserdar:
If invalid or nil key is passed, Claims() returns unmarshaled claims along with a non-nil error.

### DIFF
--- a/jwt/jwt.go
+++ b/jwt/jwt.go
@@ -18,6 +18,7 @@
 package jwt
 
 import (
+	"fmt"
 	"gopkg.in/square/go-jose.v2"
 	"gopkg.in/square/go-jose.v2/json"
 	"strings"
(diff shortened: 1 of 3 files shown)


csstaub:
This seems almost a bit too "clever", the verification key that was passed in could be `nil` by accident. Can you explain what use-case you have for reading unverified claims? I'd rather add a function along the lines of `UnsafeReadClaimsWithoutVerification` to make it clear what's happening, rather than doing the nil-key thing.

bserdar:
Note that if key is nil, an error is still returned.

My use case is when dealing with multiple token issuers, each with a different key, so when you get a token, you first get the issuer, and then find out the key for the issuer so you can validate it.

Something like ReadWithoutValidation would work as well. I worked with another JWT implementation that does what I did and I liked it, so I am biased. Let me know, and I can change the PR.

bserdar:
Any decision on this? Separate func, or return with error? I can do either way..

csstaub:
I prefer a separate function with a name that clearly indicates that it is unsafe to call.

bserdar:
Added UnsafeClaimsWithoutVerification()

Issue #182, opened 2018-05-16
Validate input to crypto/aes to avoid panic when JWT has been tampered with


mmccormack-scwx:
gopkg.in/square/go-jose.v2

In cases where the input to jwt.ParseEncrypted() may be corrupted or tampered with, the subsequent call to Claims() may panic;

panic: cipher: incorrect nonce length given to GCM [recovered]

Ideally we'd avoid a panic state in a server scenario ingesting and parsing JWT headers that may be tampered with (e.g. during pentesting).

repro:

```
key := []byte("1234567890123456")

sig, _ := jose.NewEncrypter(
  jose.A128GCM,
  jose.Recipient{Algorithm: jose.DIRECT,
    Key: key},
  (&jose.EncrypterOptions{}).WithType("JWT"))

cl := jwt.Claims{
    Subject:  "test@test.com",
    Issuer:   "repro",    
    IssuedAt: jwt.NewNumericDate(time.Now()),
  }

raw, _ := jwt.Encrypted(sig).Claims(cl).CompactSerialize()

// modify with valid base64 junk
r := strings.Split(raw, ".")
r[2] = "b3RoZXJ0aGluZw"
raw = strings.Join(r, ".")

tok, _ := jwt.ParseEncrypted(raw)

cl = jwt.Claims{}
tok.Claims(i.key, &cl)  <- AES panics
```

mmccormack-scwx:
Looks good, thanks very much.

PR #183, opened 2018-05-16
Check inputs before calling AEAD decrypt

csstaub:
Unfortunately the standard lib’s AES-GCM panics if the inputs are too short instead of returning an error, see issue #182. Will cherry-pick this fix into v1 after merge also since I’m pretty sure it has the same problem.


mcpherrinm:
lgtm

PR #186, opened 2018-06-11
Fix panic on invalid nonce in JWE

zerodivisi0n:
AES GCM `Open` method requires that the nonce should be a [strictly defined size](https://github.com/golang/go/blob/b3562658fddef6e9008379cac16c04c26784b7ed/src/crypto/aes/aes_gcm.go#L138-L140).
So it is still possible to tamper data and get panic.
This PR fixes this.

### DIFF
--- a/jwe_test.go
+++ b/jwe_test.go
@@ -23,6 +23,7 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"math/big"
+	"regexp"
 	"testing"
 )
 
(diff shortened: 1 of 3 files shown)


csstaub:
Thanks @zerodivisi0n! Merged into `v2` on `3f1dcb268fd42ad0ab2147a568b17e4db23a8eb8` through a cherry-pick (ended up cherry-picking to resolve the test file conflict).

PR #190, opened 2018-07-09
Add support for PBES2 Key algorithms

maraino:
Add functionality to support for password-based cryptography with the optional key algorithms:
 * PBES2-HS256+A128KW
 * PBES2-HS384+A192KW
 * PBES2-HS512+A256KW

PBES2 algorithms require two new parameters in the header p2c (pbkdf2 number of iterations) and p2s (salt). I added those parameters in the recipient and they are optional. Safe defaults are used if they are left empty, 100,000 for the count and a 128 bits salt. NIST recommendation is at least 10,000, but some applications like 1Password are already using 100k.

I'm open to suggestions, do a PR for the master branch, change the defaults, or use EncrypterOptions to add the extra parameters instead of the recipient.

### DIFF
--- a/crypter.go
+++ b/crypter.go
@@ -108,6 +108,8 @@ type Recipient struct {
 	Algorithm KeyAlgorithm
 	Key       interface{}
 	KeyID     string
+	P2C       int
+	P2S       []byte
 }
 
 // NewEncrypter creates an appropriate encrypter based on the key type
@@ -226,6 +228,14 @@ func (ctx *genericEncrypter) addRecipient(recipient Recipient) (err error) {
 	recipientInfo, err = makeJWERecipient(recipient.Algorithm, recipient.Key)
 	if recipient.KeyID != "" {
 		recipientInfo.keyID = recipient.KeyID
+	}
+
+	switch recipient.Algorithm {
+	case PBES2_HS256_A128KW, PBES2_HS384_A192KW, PBES2_HS512_A256KW:
+		if sr, ok := recipientInfo.keyEncrypter.(*symmetricKeyCipher); ok {
+			sr.p2c = recipient.P2C
+			sr.p2s = recipient.P2S
+		}
 	}
 
 	if err == nil {
--- a/crypter.go
+++ b/crypter.go
@@ -104,12 +104,18 @@ func (eo *EncrypterOptions) WithType(typ ContentType) *EncrypterOptions {
 }
 
 // Recipient represents an algorithm/key to encrypt messages to.
+//
+// PBES2Count and PBES2Salt correspond with the  "p2c" and "p2s" headers used
+// on the password-based encryption algorithms PBES2-HS256+A128KW,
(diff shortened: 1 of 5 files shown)


csstaub:
Hi @maraino! Thanks for your pull request! I finally have time to review this, will take a look now.

crypter.go
```diff
@@ -108,6 +108,8 @@ type Recipient struct {
 	Algorithm KeyAlgorithm
 	Key       interface{}
 	KeyID     string
+	P2C       int
+	P2S       []byte
```
csstaub: Can we change the names of these fields to make it clear what they do? Or, alternatively, add doc comments.

shared.go
```diff
@@ -320,6 +324,26 @@ func (parsed rawHeader) getCritical() ([]string, error) {
 	return q, nil
 }
 
+// getS2C extracts parsed "p2c" from the raw JSON.
+func (parsed rawHeader) getP2C() (int, error) {
+	v := parsed[headerP2C]
+	if v == nil {
+		return 0, nil
```
csstaub: I think this should probably return an error here.

csstaub:
This looks pretty good! It looks like they key is still passed in as `[]byte`, but since this is for password-based encryption maybe we have the type for the key be `string` for PBES2 algorithms? What do you think?

shared.go
maraino: I was following the behavior of similar methods on the rawHeader type, `getString()`, `getByteBuffer()`, `getJWK()`, ...

crypter.go
maraino: Alternative names could be PBES2Count and PBES2Salt, or just Count and Salt. But I rather add docs and have the field names match better the data written in the header.

csstaub: I like `PBES2Count` and `PBES2Salt`. Since this struct doesn't directly represent a header I think it's ok to to have names not match the header fields.

maraino:
> This looks pretty good! It looks like the key is still passed in as []byte, but since this is for password-based encryption maybe we have the type for the key be string for PBES2 algorithms? What do you think?
@csstaub: that makes sense. I can accept both string and []byte

crypter.go
maraino: Ok, I'll change it

maraino:
@csstaub: just pushed the changes requested. For consistency with other methods, I haven't change the `getP2C()` behavior.

maraino:
@csstaub: any update on this?

csstaub:
@maraino LGTM, thanks for your contribution! If you would like for me to tag/release a new version with this included let me know.

maraino:
>  If you would like for me to tag/release a new version with this included let me know.

@csstaub: I leave this up to you, I have no preferences.

maraino:
@csstaub: go ahead and add a new tag, it will be cleaner and easier to manage.

csstaub:
Tagged as v2.1.8: https://github.com/square/go-jose/releases/tag/v2.1.8

PR #191, opened 2018-07-14
Fix typo in comment

darkk:
I hope, `v2` is the correct branch to submit PR to as `master` is default GH branch, but README suggests `v2` :)


csstaub:
Thanks for your pull request! Yes, that's the correct branch. I should probably make that the default at some point and drop the master branch entirely.

Issue #193, opened 2018-07-25
There is no support for non native Golang keys, non standard algorithms and alternate entropy sources.


dmercer-google:
tl;dr but I am trying to make a case for why this is important.

There are many instances where the keys, algorithms and the entropy source required to sign a JWT are not supported by Golang and go-jose.  [Issue #25](https://github.com/square/go-jose/issues/25) is an example of one such case.

Some examples of unsupported key types include:
* PKCS11 keys (e.g. those provided by [miekg/pkcs11](https://github.com/miekg/pkcs11)). 
* Keys associated with Hardware Security Modules (HSM) exposing APIs other than PKCS11 (e.g. nCipher ncore keys).  
* ECC keys using [NIST P-192](https://csrc.nist.gov/csrc/media/publications/fips/186/3/archive/2009-06-25/documents/fips_186-3.pdf) (see Appendix D para D2.1)  which is unsupported by Golang.
* Key handles associated with various cloud providers (e.g. AWS, Azure and GCP)

There are instances where, so as to comply with internal security policy, signing must be performed with keys that are supported by Golang and/or go-jose but which employ the use of entropy sources other than ```crypto/rand.Reader```. An example of this is where entropy must be generated through specific interaction with a hardware based entropy source (e.g. an HSM)

go-jose only supports a subset of the signing algorithms called out in the [IANA registry of JSON Web Signature and Encryption Algorithms](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms).  Unsupported algorithms of note are ```EdDSA``` and ```none```.

There are instances where algorithms not mentioned in the [IANA registry of JSON Web Signature and Encryption Algorithms](https://www.iana.org/assignments/jose/jose.xhtml#web-signature-encryption-algorithms) are required. 

Some examples of such algorithms are:
* [GOST R 34.10-2001](https://tools.ietf.org/html/rfc5832), 
* [KCDSA](https://en.wikipedia.org/wiki/KCDSA)

[RFC 7515 para 4.1.1](https://tools.ietf.org/html/rfc7515#page-10) explicitly allows for the use of unregistered algorithms providing they do not lead to name collisions with IANA registered algorithms.

[...]

PR #194, opened 2018-07-25
Support for non native Golang keys, non standard algorithms and alternate entropy sources

dmercer-google:
This is a solution to [issue 193](https://github.com/square/go-jose/issues/193) which outlines the motivation in much more detail than I'll do here.

This PR implements the notion of an AbstractSigner and AbstractVerifier. By implementing these 
interfaces a developer can extend the key types supported for signing beyond those native to Golang
and go-jose as well as implement support for unsupported algorithms and entropy sources other than
crypto/rand.Reader.

An example of how to implement such signers/verifiers is included in the test code.

**Question:**  Should I also include something like what I have in the test code doc_test.go?

### DIFF
--- a/abstract_key_test.go
+++ b/abstract_key_test.go
@@ -1,0 +1,139 @@
+/*-
+ * Copyright 2018 Square Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package jose
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+	"testing"
+)
+
+const (
+	JwsPayload = `{"key":"value"}`
+)
+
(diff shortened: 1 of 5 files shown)


abstract_key_test.go
```diff
@@ -0,0 +1,139 @@
...
+const (
+	JwsPayload = `{"key":"value"}`
+)
+
+var Key = []byte{0x01, 0x02, 0x03, 0x04, 0x05}
+var WrongKey = []byte{0x06, 0x07, 0x08}
+
+var serialized string
+
+func TesSignWithAbstractSigner(t *testing.T) {
```
alokmenghrajani: nit: s/Tes/Test/

dmercer-google: @alokmenghrajani 

Dude, I spent like 30 min trying to figure out why I could not get this test to run in GoLand. Nice catch.  Fixed in [this commit.](https://github.com/dmercer-google/go-jose/commit/76e8c531fc7f7d186eb92564786befd698e45d5f)

csstaub:
Hi @dmercer-google, thanks for your pull request! It looks like you're trying to merge into master, which is currently still tracking `v1` (I should probably drop the master branch at some point, sorry for the confusion). For new additions we'd like to have them to into the `v2` branch. Also, we have the https://github.com/square/go-jose/tree/v2/cryptosigner package in v2, which I think might cover some of your use case? Take a look and let me know what you think.

dmercer-google:
@csstaub 

Hahaha.  One would think that one of us had copied the others code and just swapped Abstract for Opaque.

I think that OpaqueSigner/Verifier does exactly what I tried to do with this PR.  Feel free to reject this PR if you agree.

csstaub:
Sounds good, I will reject this pull request then @dmercer-google. Also having multiple engineers arrive at the same design for this feature is probably a good indication that it's a solid design!

PR #201, opened 2018-09-19
Clarification of what leeway means

bserdar:
This is a documentation change explaining what leeway really means, and how the behavior changes depending on where it is used. This caused me some trouble when I used it to validate tokens from the client side.

### DIFF
--- a/jwt/validation.go
+++ b/jwt/validation.go
@@ -47,6 +47,14 @@ func (e Expected) WithTime(t time.Time) Expected {
 
 // Validate checks claims in a token against expected values.
 // A default leeway value of one minute is used to compare time values.
+//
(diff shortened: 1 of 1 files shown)


jwt/validation.go
```diff
@@ -47,6 +47,14 @@ func (e Expected) WithTime(t time.Time) Expected {
 
 // Validate checks claims in a token against expected values.
 // A default leeway value of one minute is used to compare time values.
+//
+// The leeway will cause the token to be deemed valid until one minute
```
csstaub: Maybe write "default leeway" here to indicate that that's just the default value, not a fixed value?

csstaub:
Looks good to me! Thanks for improving the documentation @bserdar.

Issue #202, opened 2018-09-19
Include precomputed RSA values in JSONWebKey.MarshallJSON


Everlag:
Currently, go-jose doesn't copy over the pre-computed values(`dp`, `dq`, `qi`) if they exist on that key.

https://github.com/square/go-jose/blob/v2.1.8/jwk.go#L488

This results in an incompatibility with [node-jose](https://github.com/cisco/node-jose/issues/77). This is a rather difficult to debug issue given the opaque error node-jose generates, `unsupported algorithm`

I'm preparing an associated PR.

Everlag:
Resolved in https://github.com/square/go-jose/pull/203

csstaub:
Released in v2.1.9

PR #203, opened 2018-09-19
Include precomputed RSA values in JSONWebKey.MarshallJSON

Everlag:
Currently, go-jose doesn't copy over the pre-computed
values(dp, dq, qi) if they exist on that key.

https://github.com/square/go-jose/blob/v2.1.8/jwk.go#L488

This results in an incompatibility with node-jose. This is a
rather difficult to debug issue given the opaque error
node-jose generates, `unsupported algorithm`.

https://github.com/square/go-jose/issues/202

--------------

`go test` passes locally for me


Everlag:
Signed and added corporate email, waiting for propagation.

Everlag:
Thanks. I'm not sure what you mean, this change should emit the precomputed values into the JWK; `Dq` and such are already present on `rawJSONWebKey` which gets serialized as previously.

Everlag:
Tagging a release would be great!

PR #210, opened 2018-12-05
Add checks for octet length of X, Y, and D

jsha:
The JWK spec says these fields in EC keys MUST be a fixed length, but go-jose wasn't
checking their length on input. I got a report from the Let's Encrypt forums that this
caused some confusion with certain software that was generating invalid JWKs. Boulder
(using go-jose) ingested those keys just fine, but output a different (valid) encoding.

This change enforces the correct length on parsing. It also changes some `t.Errorf` to
`t.Fatalf` that I noticed during testing. The `t.Fatalf` checks are for failures that prevent
the test from meaningfully continuing.

### DIFF
--- a/jwk.go
+++ b/jwk.go
@@ -566,7 +566,17 @@ func fromEcPrivateKey(ec *ecdsa.PrivateKey) (*rawJSONWebKey, error) {
 		return nil, fmt.Errorf("square/go-jose: invalid EC private key")
 	}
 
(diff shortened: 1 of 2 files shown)


csstaub:
This looks good to me. Thank you @jsha!

csstaub:
(I'll release this in a v2.2.1 later today)

PR #212, opened 2018-12-15
Export RandReader for deterministic testing

jack-stripe:
Hello!

We're using go-jose for a new project, and it'd be quite useful to be able to stub its randomness in tests. The specific kind of test we want to enable is where a partner gives us known inputs to be JWE'd and outputs to compare the results to (which gives us a way of asserting that we're actually invoking `go-jose` correctly.)

This should be a fully backwards-compatible change, but let me know what you think!


jack-stripe:
(re: CLA, I believe my employer has reached out to Square.)

csstaub:
Change looks good to me. If you need help contacting the right person about setting up the CLA just shoot me an email, I’m cs(at)squareup.com.

csstaub:
Merged, thanks for the contribution @jack-stripe! If you need me to tag this in a release let me know.

csstaub:
This is out, in release v2.2.2: https://github.com/square/go-jose/releases/tag/v2.2.2

Issue #216, opened 2019-02-02
Reject tokens issued in the future (Issued At/iat validation)


evanj:
The function `jwt.ValidateWithLeeway` checks the times on the `nbf` and `exp` claims. It does not check the `iat` claim. The standard [RFC 7519 states that this field is *optional*, and *does not require* that the time on this field be checked](https://tools.ietf.org/html/rfc7519#section-4.1.6). However, if it is issued in the future, something suspicious is happening. It seems to me that it would be good to be paranoid?

If this seems acceptable, I'm happy to attempt to submit a pull request for this, since it will only require a few lines of code to be added here: https://github.com/square/go-jose/blob/v2/jwt/validation.go#L97

Found while checking an implementation of validation for Google's Identity-Aware Proxy, which includes a test case for this: https://cloud.google.com/iap/docs/special-urls-and-headers-howto#testing_jwt_verification

csstaub:
Agree, I prefer the library being strict rather than lenient on checks, we should validate `iat` if it’s present in the JWT to ensure it’s not issued in the future (probably also applying the leeway since clocks can be slightly out of sync sometimes). Happy to accept a pull request for this!

PR #217, opened 2019-02-03
jwt.ValidateWithLeeway: Return error if IssuedAt (iat) is in the future

evanj:
Add a check to ValidateWithLeeway that the IssuedAt (iat) field is in the
past. This field is optional, and this check is not required by the RFC,
but seems like a reasonable check. See:
https://tools.ietf.org/html/rfc7519#section-4.1.6

Fixes https://github.com/square/go-jose/issues/216


csstaub:
Thanks @evanj! I'll let this sit for a bit and then make a release with the change.

PR #218, opened 2019-02-07
Allow JSONWebKeySet for retrieving claims and decrypting nested token

shaxbee:


jwt/jwt.go
```diff
@@ -130,3 +135,29 @@ func ParseSignedAndEncrypted(s string) (*NestedJSONWebToken, error) {
...
+	if kid == "" {
+		return
+	}
+
+	keys := jwks.Key(kid)
+	if len(keys) == 0 {
+		return
+	}
+
+	*key = keys[0].Key
```
csstaub: Would prefer if this returned the key instead of updating via the pointer. Not clear from function signature that it has side-effects.

csstaub:
Updated this in f1cd7546ee6ddbba796a5e88528966689a186a76 and merged. Thanks @shaxbee!

shaxbee:
Cheers!

Issue #219, opened 2019-02-23
Allow tokens that do not have expiry set


nelz9999:
Expiry "exp" is an OPTIONAL claim, as per the spec: https://tools.ietf.org/html/rfc7519#section-4.1.4

Expiry should not be a failing case if unset, even if the Expected.Time is non-zero.

Here's a unit test to show what's wrong:
```
package main

import (
	"testing"
	"time"

	"gopkg.in/square/go-jose.v2/jwt"
)

func TestJWTValidateExpired(t *testing.T) {
	testCases := []struct {
		name string
		body jwt.Claims
		fail bool
	}{
		{
			"future expiration",
			jwt.Claims{
				Expiry: jwt.NewNumericDate(time.Now().Add(time.Hour)),
			},
			false,
		},
		{
			"past expiration",
			jwt.Claims{
				Expiry: jwt.NewNumericDate(time.Now().Add(-time.Hour)),
			},
			true,
		},
		{
			"no expiry",
			jwt.Claims{},
			false,
		},
	}

	for _, tc := range testCases {
		t.Run(tc.name, func(t *testing.T) {
			err := tc.body.Validate(jwt.Expected{}.WithTime(time.Now()))
			if err != nil && !tc.fail {
				t.Errorf("unexpected: %v\n", err)
			}
			if err == nil && tc.fail {
				t.Errorf("did not receive expected failure\n")
			}
		})
	}
}
```

And the output of the test:
```
=== RUN   TestJWTValidateExpired
=== RUN   TestJWTValidateExpired/future_expiration
=== RUN   TestJWTValidateExpired/past_expiration
=== RUN   TestJWTValidateExpired/no_expiry
--- FAIL: TestJWTValidateExpired (0.00s)
    --- PASS: TestJWTValidateExpired/future_expiration (0.00s)
    --- PASS: TestJWTValidateExpired/past_expiration (0.00s)
    --- FAIL: TestJWTValidateExpired/no_expiry (0.00s)
        main_test.go:41: unexpected: square/go-jose/jwt: validation failed, token is expired (exp)
FAIL
coverage: 0.0% of statements
exit status 1
FAIL	github.com/nelz9999/scratchpad/jose	0.028s
```

In `jwt/validation.go`, I believe this line:
```
    if !e.Time.IsZero() && e.Time.Add(-leeway).After(c.Expiry.Time()) {
```

Should be:
```
[...]

PR #220, opened 2019-02-25
optional date claims

nelz9999:
Solves #219

### DIFF
--- a/jwt/validation.go
+++ b/jwt/validation.go
@@ -94,17 +94,17 @@ func (c Claims) ValidateWithLeeway(e Expected, leeway time.Duration) error {
 		}
 	}
 
-	if !e.Time.IsZero() && e.Time.Add(leeway).Before(c.NotBefore.Time()) {
+	if c.NotBefore != nil && !e.Time.IsZero() && e.Time.Add(leeway).Before(c.NotBefore.Time()) {
 		return ErrNotValidYet
 	}
 
-	if !e.Time.IsZero() && e.Time.Add(-leeway).After(c.Expiry.Time()) {
+	if c.Expiry != nil && !e.Time.IsZero() && e.Time.Add(-leeway).After(c.Expiry.Time()) {
 		return ErrExpired
 	}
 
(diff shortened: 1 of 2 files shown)


nelz9999:
Aww jeez, `t.Run(...)` doesn't exist pre-1.7. 😢 I'll have to tidy that up later tonight.

mcpherrinm:
1.7 is pretty old at this point.  Seems fine to just bump the requirements up to something more recent.

mcpherrinm:
sorry, I accidentally pressed "close and comment" instead of just "comment"

nelz9999:
Bumping to 1.7 seems like a bigger breaking change than I feel empowered to make. I'll probably just make my tests work in pre-1.7, and let the low-version policy decision be handled elsewhere.

csstaub:
Go 1.7 came out in August 2016, seems reasonable to bump that requirement. 

Here's a PR for it to rebase on: https://github.com/square/go-jose/pull/221

nelz9999:
Rebased, and it seems to have passed!

csstaub:
This looks good, thank you @nelz9999!

One concern I have that I'd like input on: some callers might want to enforce the presence of an expiration in the token. This was previously enforced (though not on purpose I suppose) but now isn't. It's probably sufficient to make this clear in the release notes, but maybe this is something we'd want to support in the API. Any thoughts or opinions on this?

nelz9999:
> some callers might want to enforce the presence of an expiration in the token

This is a little related to the scenario I'm programming against right now. My feeling is that much like non-standard claims, if the caller wants to verify things surpassing the requirements of the spec, they should do that on their own. And I believe the `JSONWebToken.Claims(...)` decoding into arbitrary structs is a fine affordance for that.

nelz9999:
And now I'm gonna be a little annoying... D'you have any idea when this might be released? 😇

csstaub:
Soon, I just have to type up release notes and such

csstaub:
Voilà: https://github.com/square/go-jose/releases/tag/v2.3.0

PR #224, opened 2019-04-05
Reverse bytes in Ed25519 keys.

maraino:
Golang golang.org/x/crypto/ed25519 package uses 64 bytes to store
the public and private parts of the key, the first 32 bytes correspond
to the private part and the last 32 to the public part. See:
https://github.com/golang/crypto/blob/master/ed25519/ed25519.go#L90

go-jose implementation reverses the bytes causing that the JSON version
of the private key has the private part in the 'x' property and the
public in the 'd', and it should be the other way around, see:
https://tools.ietf.org/html/rfc8037#appendix-A.1

### DIFF
--- a/jwk.go
+++ b/jwk.go
@@ -230,7 +230,7 @@ func (k *JSONWebKey) Thumbprint(hash crypto.Hash) ([]byte, error) {
 	case *rsa.PrivateKey:
 		input, err = rsaThumbprintInput(key.N, key.E)
 	case ed25519.PrivateKey:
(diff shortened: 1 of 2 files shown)


csstaub:
Wow, that's a really good catch, thanks @maraino.

maraino:
Thanks @csstaub, can you tag a new release?

csstaub:
@maraino Already on it!

csstaub:
Release in v2.3.1: https://github.com/square/go-jose/releases/tag/v2.3.1

PR #227, opened 2019-04-16
If KeyId is empty, do not include it in the header

paulquerna-okta:
The current implementation can end up with an empty `kid` field in the JWT header:

```
{
  "alg": "RS256",
  "kid": ""
}
```

In other places, `kid` is treated as an `,omitempty` when using `encoding/json` struct tags, but in this code path we are building up the map ourselves, and therefore we need to check if `.KeyID` is empty directly.


csstaub:
This LGTM, will merge this into v2. Thank you for your contribution @paulquerna-okta! Do you need me to tag a release with this change?

Issue #223, opened 2019-04-03
Go modules support


philtay:
what about this plan?
* v2 branch should be merged back to master
* add `go.mod` and `go.sum` files in the repository root (master)
* cut new release

mbyczkowski:
I think it would be great to eventually move go-jose to use Go modules and just use master branch for development (as opposed to `v1` and `v2`). However we want to be careful, so people using old imports (like `gopkg.in/square/go-jose.v2`) still get the updates without making code changes.

This could be a good time to create a `v3` release and figure out support plan for `v1` and `v2`(would they still be actively maintained? receive security fixes only?).

Since we tag releases, go-jose as a lib already works pretty well for people using Go modules, so there's no rush here to migrate. I think Go modules would be more useful for tools that we include in this repo (like `jose-util`). Should those programs be separate modules though (or be moved to a different repo)? I don't know if we have an answer to that, yet.

csstaub:
I agree with @mbyczkowski's assessment, since Go modules support gopkg.in import paths go-jose should already work for people using Go modules for their code. If we were to switch to Go modules for the repo we'd probably do it in a v3 release but I don't see a benefit to that, at the moment.

shaxbee:
Go module support would enable caching go-jose in proxy (such as gocenter.io)
dramatically reducing build times which is crucial for CI.

csstaub:
That's good to know, I wasn't aware of that. Maybe makes sense to plan a v3 then.

PR #230, opened 2019-05-13
Support the b64 header parameter

### LINKED ISSUE #225: Support the "b64" header parameter
https://tools.ietf.org/html/rfc7797#section-3

Currently there is no way to sign the payload *not* encoded in base64. That's a serious limitation when the service you're sending the signatures to requires `b64=false`. If `SignerOptions.ExtraHeaders` has `b64` key with a `false` value then the payload should not be encoded.

philtay:
Closes #225

### DIFF
--- a/jws.go
+++ b/jws.go
@@ -95,15 +95,12 @@ func (sig Signature) mergedHeaders() rawHeader {
 
 // Compute data to be signed
 func (obj JSONWebSignature) computeAuthData(payload []byte, signature *Signature) []byte {
-	var (
-		buf bytes.Buffer
-		enc = true
-		err error
-		pro = new(rawHeader)
-	)
+	var buf bytes.Buffer
+
+	pro := new(rawHeader)
 
 	if signature.original != nil && signature.original.Protected != nil {
-		if err = json.Unmarshal(signature.original.Protected.bytes(), pro); err != nil {
+		if err := json.Unmarshal(signature.original.Protected.bytes(), pro); err != nil {
 			panic(err)
 		}
 		buf.WriteString(signature.original.Protected.base64())
@@ -112,7 +109,10 @@ func (obj JSONWebSignature) computeAuthData(payload []byte, signature *Signature
 		buf.WriteString(base64.RawURLEncoding.EncodeToString(mustSerializeJSON(pro)))
 	}
 
+	enc := true
+
 	if pro != nil {
+		var err error
 		if enc, err = pro.getB64(); err != nil {
 			enc = true
 		}
--- a/jws.go
+++ b/jws.go
@@ -95,38 +95,38 @@ func (sig Signature) mergedHeaders() rawHeader {
 
 // Compute data to be signed
 func (obj JSONWebSignature) computeAuthData(payload []byte, signature *Signature) []byte {
-	var buf bytes.Buffer
+	var authData bytes.Buffer
 
-	pro := new(rawHeader)
+	protectedHeader := new(rawHeader)
 
 	if signature.original != nil && signature.original.Protected != nil {
-		if err := json.Unmarshal(signature.original.Protected.bytes(), pro); err != nil {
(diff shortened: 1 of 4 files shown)


philtay:
This PR is ready for review.

/cc @csstaub @mcpherrinm

jws.go
```diff
@@ -94,19 +95,38 @@ func (sig Signature) mergedHeaders() rawHeader {
 
 // Compute data to be signed
 func (obj JSONWebSignature) computeAuthData(payload []byte, signature *Signature) []byte {
-	var serializedProtected string
+	var buf bytes.Buffer
+
+	pro := new(rawHeader)
```
csstaub: This is a nit, but I'd prefer if we gave these variables more descriptive names. It's not clear at first glance what "pro", "enc" and "buf" mean or what they're used for.

csstaub: How about calling these protectedHeader, needsBase64, and authData (or something along those lines)?

shared.go
```diff
@@ -349,6 +350,21 @@ func (parsed rawHeader) getP2S() (*byteBuffer, error) {
 	return parsed.getByteBuffer(headerP2S)
 }
 
+// getB64 extracts parsed "b64" from the raw JSON, defaulting to true.
+func (parsed rawHeader) getB64() (bool, error) {
```
csstaub: Would it ever be necessary here to distinguish between the header being present, and set, or the header being not present and the default return value being false?

signing.go
```diff
@@ -254,11 +255,26 @@ func (ctx *genericSigner) Sign(payload []byte) (*JSONWebSignature, error) {
...
-		input := []byte(fmt.Sprintf("%s.%s",
-			base64.RawURLEncoding.EncodeToString(serializedProtected),
-			base64.RawURLEncoding.EncodeToString(payload)))
+		var input bytes.Buffer
 
-		signatureInfo, err := recipient.signer.signPayload(input, recipient.sigAlg)
+		input.WriteString(base64.RawURLEncoding.EncodeToString(serializedProtected))
+		input.WriteByte('.')
+
+		enc := true
```
csstaub: Same comment as above, let's call this something other than "enc". Could be confused to mean "encrypt" for example.

signing_test.go
```diff
@@ -521,3 +521,39 @@ func TestSignerExtraHeaderInclusion(t *testing.T) {
...
 }
+
+func TestSignerB64(t *testing.T) {
+	signingKey, verificationKey := GenerateSigningTestKey(RS256)
+
+	opts := &SignerOptions{
+		ExtraHeaders: map[HeaderKey]interface{}{
+			"b64": false,
+		},
+	}
```
csstaub: This seems somewhat non-obvious, would it be better to have this be a first-class option in SignerOptions and have the signing code add the header if that's selected?

csstaub:
Hi @philtay, thank you so much for your pull request! I left a couple of comments inline, but I think what's also missing is code to add the b64 header to the list of crit headers (see RFC 7797, Section 6: https://tools.ietf.org/html/rfc7797#section-6).

philtay:
@csstaub thanks for the review. The last commit takes in account your comments and brings some improvements.

philtay:
@csstaub can you review the updated PR? I'm eager to have this merged. Thanks!

csstaub:
Hi @philtay, changes look good to me! Thank you very much for your contribution. I'll merge this and play around with it a bit and then will try to make a tagged release later this week.

philtay:
@csstaub thanks! I've another PR to open before making a tagged release. A `func (obj JSONWebSignature) DetachedCompactSerialize() (string, error)` method as per RFC 7797 (compact serialization and detached payload). I'll open the PR later today or tomorrow. If you want, I'm also interested in closing #223 and cutting a v3.0.0. Just merge the `v2` branch  into `master` and I'll do the rest. Let me know.

csstaub:
Sounds good, let's do it.

PR #235, opened 2019-05-22
Drop support for old Go releases

philtay:
See https://github.com/square/go-jose/pull/231#pullrequestreview-236801235

/cc @csstaub @mcpherrinm


csstaub:
Thanks @philtay!

PR #238, opened 2019-05-24
Changes for version 3 of jose-util

csstaub:
Changes:
* Massive code cleanup, split code from one file into smaller files
* Added base64-decode command that handles url-safe base64 that is used in JOSE messages
* Moved jwk-keygen functionality into jose-util as a subcommand


philtay:
A more common pattern would be to put this command in a `/cmd` subdirectory. Then you could also rename it `jose` instead of `jose-util`. I also suggest to have a single `go.mod` and `go.sum` file in the repo root and to avoid vendoring.

csstaub:
I'd prefer to keep the modules separate, I wouldn't want dependencies of the util to become dependencies of the module.

Issue #240, opened 2019-05-28
Add support for x5u field in rawJSONWebKey struct


0x13a:
I'd like to parse a json web key which has a `x5u` field as compliant with the rfc spec https://tools.ietf.org/html/rfc7517#section-4.6

However looks like this specific field is not available in the struct

```go
type rawJSONWebKey struct {
	Use string      `json:"use,omitempty"`
	Kty string      `json:"kty,omitempty"`
	Kid string      `json:"kid,omitempty"`
	Crv string      `json:"crv,omitempty"`
	Alg string      `json:"alg,omitempty"`
	K   *byteBuffer `json:"k,omitempty"`
	X   *byteBuffer `json:"x,omitempty"`
	Y   *byteBuffer `json:"y,omitempty"`
	N   *byteBuffer `json:"n,omitempty"`
	E   *byteBuffer `json:"e,omitempty"`
	// -- Following fields are only used for private keys --
	// RSA uses D, P and Q, while ECDSA uses only D. Fields Dp, Dq, and Qi are
	// completely optional. Therefore for RSA/ECDSA, D != nil is a contract that
	// we have a private key whereas D == nil means we have only a public key.
	D  *byteBuffer `json:"d,omitempty"`
	P  *byteBuffer `json:"p,omitempty"`
	Q  *byteBuffer `json:"q,omitempty"`
	Dp *byteBuffer `json:"dp,omitempty"`
	Dq *byteBuffer `json:"dq,omitempty"`
	Qi *byteBuffer `json:"qi,omitempty"`
	// Certificates
	X5c []string `json:"x5c,omitempty"`
}
```

Is there something I am missing or is the x5u field really not being parsed?

Regards

csstaub:
It's not populated in the struct at the moment. But it should be easy to add if you need it, please open a pull request against `master` (we're currently prepping that for v3 release).

0x13a:
merged https://github.com/square/go-jose/pull/242 - thanks @csstaub

PR #241, opened 2019-05-28
Add support for x5u parameter in jwk

0x13a:
as discussed here https://github.com/square/go-jose/issues/240

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,4 +2,7 @@ module github.com/square/go-jose
 
 go 1.12
 
-require golang.org/x/crypto v0.0.0-20190513172903-22d7a77e9e5f
+require (
+	github.com/stretchr/testify v1.3.0
+	golang.org/x/crypto v0.0.0-20190513172903-22d7a77e9e5f
+)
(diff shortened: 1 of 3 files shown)


csstaub:
This won't expose it to callers of the library though, which is probably what you want (the raw representation is private and is only for a pre-parsing step). Let me work on a quick PR to add this, I realized while we're at it we should also add x5t and x5t#S256 probably and validate their values.

0x13a:
yes thanks. would certificate URI suffice and makes sense? please check commit 
e76ad40

0x13a:
@csstaub I can also add the thumbprint and the sha256-Thumbprint parameters as well. do you want to do some sort of validation during unmarshalling?

csstaub:
Yes, I have some code changes I'm working on. Hang tight.

csstaub:
See #242 for a proposed implementation with x5u/x5t support and additional validation checks.

0x13a:
brilliant, thanks

PR #242, opened 2019-05-28
Support x5u, x5t, and x5t#S256 headers (for JWK)

csstaub:
Changes:
* Support x5u, x5t, and x5t#S256 headers.
* Check that cert pub key matches JWK pub key on unmarshal.
* Check that thumbprints match X.509 leaf cert on marshal/unmarshal.
* Add additional tests for invalid thumbprints, cert/key mismatch.

Note that we _could_ do additional checks on the cert chain, like compare the key usage extension to the defined "use" in the JWK header. But I decided not to implement that at this time -- the standard is a bit fuzzy about how to match these (sigh). We also don't check the cert chain that the x5u URI points at, because we don't want to do an HTTP get on a JSON unmarshal and it could change at any time regardless.

cc @0x13a

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -2,4 +2,7 @@ module github.com/square/go-jose
 
 go 1.12
 
-require golang.org/x/crypto v0.0.0-20190513172903-22d7a77e9e5f
+require (
+	github.com/stretchr/testify v1.3.0
+	golang.org/x/crypto v0.0.0-20190513172903-22d7a77e9e5f
+)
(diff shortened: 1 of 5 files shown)


ramonberrutti:
@csstaub x5u need to be a string? WDYT?

mbyczkowski:
@ramonberrutti if you think there might be a bug, just open an issue and we can discuss there. Not many people will look at already-merged PRs.

PR #244, opened 2019-05-30
Third lint pass

philtay:
This PR also includes GolangCI-Lint in Travis support.


utils_test.go
```diff
@@ -26,7 +26,7 @@ import (
 
 // Reset random reader to original value
 func resetRandReader() {
-	RandReader = rand.Reader
+	randReader = rand.Reader
```
csstaub: Let's keep this one public. It was exposed on purpose to make it possible to test against deterministic test vectors in client libraries (which I think is a reasonable request/thing to do).

.golangci.yml
```diff
@@ -0,0 +1,49 @@
...
+        - golint
+        - govet
+        - ineffassign
+        - staticcheck
+        - structcheck
+        - stylecheck
+        - unused
+    - path: _test\.go
+      linters:
+        - scopelint
```
csstaub: Nice!

philtay:
@csstaub done

csstaub:
Thank you @philtay!

Issue #228, opened 2019-04-16
jwe using Curve P521, ECDH-ES, and A256GCM does not always work with with other implementations


dodgeware:
I am not able to encrypt/decrypt within Java and Go using Curve P521, ECDH-ES, and A256GCM using the same values for curve X, Y, and D. When I try to take the encrypted value from Java and Decrypt in Go it fails (about half the time) and vice versa it fails with:

Go: square/go-jose: error in cryptographic primitive

Sometimes it works though.

I am manually creating the KeyPair in both programs with the same X, Y, and D.  Here is a link to my stackoverflow question.

https://stackoverflow.com/questions/55711277/why-do-i-get-an-error-decrypting-jwe-between-java-and-go

csstaub:
Can you give us more information to help debug?
* What version of Go and go-jose are you using?
* What library are you using in Java to perform the JWE operations?
* Do you have example messages/ciphertexts to test?

dodgeware:
go version: go version go1.11.8 darwin/amd64
go-jose version from go mod: gopkg.in/square/go-jose.v2 v2.3.1

Java Libraries POM:

		<dependency>
			<groupId>com.nimbusds</groupId>
			<artifactId>nimbus-jose-jwt</artifactId>
			<version>7.0</version>
		</dependency>
		<dependency>
			<groupId>org.bouncycastle</groupId>
			<artifactId>bcprov-jdk15on</artifactId>
			<version>1.61</version>
		</dependency>

JWK:
{  
   "kty":"EC",
   "d":"AGPoHUdXajyyRLV0bAokQTnDzlO7Kjs1zSucSu69CGfSwpg7oXSxlfptApD-5O47d1PX3y0ag5228XsPFXVzYnH0",
   "crv":"P-521",
   "x":"AVuFsno89wJ5xT2z63iznxVO8H5gsfcHmS1XJ_JbfEzIsudqjrvKGrzxJT96-dmP_NY7KeMvyJEUInmqcqCWbzcQ",
   "y":"ANv5hayQ3_TwMcFqPrtw-a9wNkfQuynuWhhbWOXYvGArdvibDGYRIRx3O5gAjfyumpibZFQ0K0jrjb09YP3AVbtc"
}

Encryption Text:  "test string"

Sample Encryption from Java (copied into Go Program and attempt decryption): 

[...]

PR #245, opened 2019-05-31
Pad z value to proper size after P-521 scalar multiplication.

csstaub:
Fixes issue #228. After calling ScalarMult for P-521, the output can sometimes be 65 bytes long instead of 66 bytes. This happens when the first bit of the computed value is zero. Calling z.Bytes() on the big integer will then omit the leading zero, giving us a 65-byte value.

This subsequently causes the shared secret computation to be incorrect as the input into the KDF function should always be 66 bytes which is the full length for a P-521 coordinate value.

### DIFF
--- a/cipher/ecdh_es.go
+++ b/cipher/ecdh_es.go
@@ -17,8 +17,10 @@
 package josecipher
 
 import (
+	"bytes"
 	"crypto"
 	"crypto/ecdsa"
+	"crypto/elliptic"
 	"encoding/binary"
 )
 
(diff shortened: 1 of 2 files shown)


csstaub:
Will cherry-pick this into `v2` after merge to master, seems important enough to fix there too.

cipher/ecdh_es.go
```diff
@@ -44,16 +46,34 @@ func DeriveECDHES(alg string, apuData, apvData []byte, priv *ecdsa.PrivateKey, p
...
 	}
 
-	z, _ := priv.PublicKey.Curve.ScalarMult(pub.X, pub.Y, priv.D.Bytes())
-	reader := NewConcatKDF(crypto.SHA256, z.Bytes(), algID, ptyUInfo, ptyVInfo, supPubInfo, []byte{})
+	z, _ := priv.Curve.ScalarMult(pub.X, pub.Y, priv.D.Bytes())
+	zBytes := z.Bytes()
 
+	octSize := dSize(priv.Curve)
+	if len(zBytes) != octSize {
+		zBytes = append(bytes.Repeat([]byte{0}, octSize-len(zBytes)), zBytes...)
```
alokmenghrajani: do we care about any timing info leak here?

cipher/ecdh_es.go
```diff
@@ -44,16 +46,34 @@ func DeriveECDHES(alg string, apuData, apvData []byte, priv *ecdsa.PrivateKey, p
 		panic("public key not on same curve as private key")
 	}
 
-	z, _ := priv.PublicKey.Curve.ScalarMult(pub.X, pub.Y, priv.D.Bytes())
-	reader := NewConcatKDF(crypto.SHA256, z.Bytes(), algID, ptyUInfo, ptyVInfo, supPubInfo, []byte{})
+	z, _ := priv.Curve.ScalarMult(pub.X, pub.Y, priv.D.Bytes())
+	zBytes := z.Bytes()
 
+	octSize := dSize(priv.Curve)
```
mbyczkowski: I think the comment from https://github.com/square/go-jose/issues/228#issuecomment-497873584 would be great here

cipher/ecdh_es.go
csstaub: Possibly, though I’m not sure how we’d fix it, and regardless P-521 in Go isn’t constant time anyway (only P-256 is afaik).

PR #247, opened 2019-06-03
Allow single string audience

dbramwell:
When trying to use this library to authenticate with [box](https://developer.box.com/docs/construct-jwt-claim-manually) I get an error stating "Please check the 'aud' claim. Should be a string".

This appears to be because the 'aud' is sent as an array rather than a string (I do think that the box api should be able to handle an array of length 1, so I've raised an issue with them as well).

According to the [spec](https://tools.ietf.org/html/rfc7519#section-4.1.3), "In the special case when the JWT has one audience, the "aud" value MAY be a single case-sensitive string containing a StringOrURI value."

The wording on this isn't super clear, and I'm inexperienced in both Go and JWT, but the way I read the above makes me think it's safe to always just send a string, rather than array, when there is only one Audience. That's what this pull request does.

### DIFF
--- a/jwt/claims.go
+++ b/jwt/claims.go
@@ -110,6 +110,16 @@ func (s *Audience) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// MarshalJSON converts audience to json representation.
+func (s Audience) MarshalJSON() ([]byte, error) {
+	type Alias Audience
+	if len(s) == 1 {
+		return json.Marshal(s[0])
+	}
+	return json.Marshal((Alias)(s))
+}
+
+//Contains checks whether a given string is included in the Audience
 func (s Audience) Contains(v string) bool {
 	for _, a := range s {
 		if a == v {
--- a/jwt/claims_test.go
+++ b/jwt/claims_test.go
@@ -45,6 +45,25 @@ func TestEncodeClaims(t *testing.T) {
 	assert.Equal(t, expected, string(b))
 }
 
+func TestEncodeClaimsWithSingleAudience(t *testing.T) {
+	now := time.Date(2016, 1, 1, 0, 0, 0, 0, time.UTC)
+
+	c := Claims{
+		Issuer:    "issuer",
+		Subject:   "subject",
+		Audience:  Audience{"a1"},
+		NotBefore: NewNumericDate(time.Time{}),
+		IssuedAt:  NewNumericDate(now),
+		Expiry:    NewNumericDate(now.Add(1 * time.Hour)),
+	}
+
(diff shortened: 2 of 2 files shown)


shaxbee:
It's working according to spec. JWT parser should accept both string and an array value and it does in case of this library. Box should be fixed instead to adhere to spec.

dbramwell:
I think the spec is a little bit ambiguous. The Auth0 Java implementation changes an array of one element to a string:
https://github.com/auth0/java-jwt/blob/8ca949eebe0d186567be72343c6e548ade001bd6/lib/src/main/java/com/auth0/jwt/impl/PayloadSerializer.java#L34

csstaub:
Huh, this is a bit of an odd one. It's like the hidden complexity in JOSE never stops 😐. 

One the one hand, going from a single-item list to a string seems reasonable if it helps with compatibility. On the other hand, we don't have a good way to check if this would work with other consumers so we might break other users of the library. Maybe we should survey and look at what other libraries do? Is a string more commonly used than a single-item list? If so, it would make sense for us to switch to that also.

csstaub:
It seems that:
* dgrijalva/jwt-go always uses a single string
* nimbus-jose-jwt flattens a single-item list to a string

Seems reasonable for us to do it as well then.

jwt/claims.go
```diff
@@ -110,6 +110,16 @@ func (s *Audience) UnmarshalJSON(b []byte) error {
 	return nil
 }
 
+// MarshalJSON converts audience to json representation.
+func (s Audience) MarshalJSON() ([]byte, error) {
+	type Alias Audience
+	if len(s) == 1 {
+		return json.Marshal(s[0])
+	}
+	return json.Marshal((Alias)(s))
```
csstaub: Since audience is already a type alias for list of string, I believe we can call into the "original" MarshalJSON here without using the type alias and just casting to []string instead.

csstaub: Like this: https://play.golang.org/p/rhQ8o6hbFJd

csstaub:
Rebased & merged in 0a88056c80b04336ce957271dd969a486e7fc951

jwt/claims.go
dbramwell: Thanks! Will give it a go tomorrow. This is my first Go go, so all feedback is welcome.

csstaub:
You can use this right now if you want by importing the master branch: 
```
import "github.com/square/go-jose"
```

But note the master branch is unstable right now. We plan to release it as v3 soon-ish (next couple of weeks is my hope) and then we'll have some stability guarantees, but until then things can still change.

PR #248, opened 2019-06-13
Expose payload verifier in interface

csstaub:
Makes the previously private payloadVerifier interface public as just a Verifier, and merges it with OpaqueVerifier. Adds type-safe functions for instantiating new verifiers from RSA, ECDSA, etc. keys so we can avoid some of the reflection we used to do internally. Should also make the interface friendlier to consumers of the library.

### DIFF
--- a/signing.go
+++ b/signing.go
@@ -34,6 +34,14 @@ type NonceSource interface {
 	Nonce() (string, error)
 }
 
+// Verifier represents logic, based on a cryptographic key, to verify a signed message.
+type Verifier interface {
+	// Verify an individual payload with a signature and algorithm. If you're dealing with
+	// a full message, you probably want to call JSONWebSignature#Verify instead of calling
+	// this directly.
+	VerifyPayload(payload []byte, signature []byte, alg SignatureAlgorithm) error
+}
+
 // Signer represents a signer which takes a payload and produces a signed JWS object.
 type Signer interface {
 	Sign(payload []byte) (*JSONWebSignature, error)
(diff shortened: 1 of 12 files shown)


signing.go
```diff
@@ -151,34 +155,51 @@ func NewMultiSigner(sigs []SigningKey, opts *SignerOptions) (Signer, error) {
 	return signer, nil
 }
 
-// newVerifier creates a verifier based on the key type
-func newVerifier(verificationKey interface{}) (payloadVerifier, error) {
-	switch verificationKey := verificationKey.(type) {
+func NewRSAVerifier(key *rsa.PublicKey) (Verifier, error) {
```
mbyczkowski: Go's idiomatic way is to _"accept interfaces, return structs"_. In this case the `Verifier` interface masks the concrete type, which is unexported. I think it'd be better to return `*RSAEncrypterVerifier` (and similar for other initializers below).

mbyczkowski: Also, this might be relevant here https://github.com/golang/go/wiki/CodeReviewComments#interfaces

csstaub: Hm, that's a very good point, but sort of makes it a larger refactor. I'd want to split those types out into a separate package. And then maybe we don't even need a NewXYZ function for these, and they can just be instantiated via the struct directly? Kinda not sure that's worth it then to refactor this way 🤔 Maybe we should hold onto this idea for v4 and just release v3 now with module support... Bumping versions later should be easy anyway so that's probably fine, if we still want this later we just do v4.

Issue #246, opened 2019-06-01
Proposal: Use semantic versioning


mbyczkowski:
As we're adopting Go modules with the upcoming version 3 of go-jose, it might be a good time to adopt semantic versioning (specifically https://semver.org/spec/v2.0.0.html).

Benefits:
- Semver clearly communicate intent with each release (especially with regards to breaking changes)
- Go modules utilizes semver already ([more here](https://github.com/golang/go/wiki/Modules#semantic-import-versioning)), so this will work well for every Go module user
- Major version change will mean updating import paths (that's how Go modules work); patch and minor version change should be safe from the API compatibility perspective - should be encouraging for users to update their version of go-jose.
- Given the work already started on V3, it should be no work to adopt semver (see [releasing Go modules with V2 or higher](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher))

If we decide to adopt it, (I think) this practically means that:
- We'll have to be more strict about making our releases (i.e. making sure we adhere to the semver spec)
- We'll state expectations with regard version support and deprecations. I think going forward we should support latest major release version + 1.

Bonus, if we can provide a note on migration from V1 and V2 to V3.

Thoughts @csstaub?

philtay:
> Major version change will mean updating import paths (that's how Go modules work)

AFAIK using Go modules + semver means to never have to change import paths again. The version is specified in the `go.mod` file and it's independent from the path.

csstaub:
I agree with the proposal @mbyczkowski.

@philtay: I think Mat B. is right here, if I understand https://github.com/golang/go/wiki/Modules#semantic-import-versioning correctly the version will need to be part of the import path like `import "github.com/square/go-jose/v3"`. So if we'd want to bump to v4, it would change to `import "github.com/square/go-jose/v4"`. Right?

mbyczkowski:
Correct, the wiki page says:
> If the module is version v2 or higher, the major version of the module _must_ be included as a `/vN` at the end of the module paths used in go.mod files (e.g., `module github.com/my/mod/v2`, `require github.com/my/mod/v2 v2.0.0`) and in the package import path (e.g., `import "github.com/my/mod/v2/mypkg"`).

> If the module is version v0 or v1, do _not_ include the major version in either the module path or the import path.

So if a library is unversioned, unstable or on 1.x release, the import path would not include the major version. V2 and higher means manually updating import paths.

[...]

PR #249, opened 2019-06-13
Update import paths for upcoming V3 release

mbyczkowski:
According to the Go modules [wiki
page](https://github.com/golang/go/wiki/Modules#releasing-modules-v2-or-higher),
both the module and import paths in the repo need to include the major
version for V2+ releases.

More details in discussion in #246

### DIFF
--- a/asymmetric.go
+++ b/asymmetric.go
@@ -28,8 +28,8 @@ import (
 	"fmt"
 	"math/big"
 
-	josecipher "github.com/square/go-jose/cipher"
-	"github.com/square/go-jose/json"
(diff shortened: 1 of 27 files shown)


mbyczkowski:
Looks like CI is failing because the V3 module is not yet available (but will be after this PR gets merged):
```
go: downloading github.com/square/go-jose/v3 v3.0.0-20190604210802-369111a8a36e
build github.com/square/go-jose/jose-util: cannot find module for path github.com/square/go-jose/v3
The command "cd jose-util && go build && PATH=$PWD:$PATH cram -v jose-util.t" exited with 1.
```

So I'm going to go ahead and merge this change.

PR #250, opened 2019-06-14
added error checking avoid nil referencing if error occurs

trung:


csstaub:
Hi @trung, looks good. Thanks for catching this and submitting a bug fix!

PR #252, opened 2019-06-26
Fix error message for incorrect pubkey x/y length.

jsha:
In #210 I introduced some error messages for incorrect octet string
length for X and Y. The error message for the public key case
incorrectly stated that it was for a private key.


csstaub:
Thanks for the fix @jsha!

PR #259, opened 2019-09-15
Point jose-util at updated package.

kousu:
Otherwise, it gets confused and downloads the version from github every time.


csstaub:
@mcpherrinm @mbyczkowski Can we turn off GolangCI (or make it non-blocking) for this project? Looks like it's failing on all the pull requests right now.

mbyczkowski:
@csstaub cool, made GolangCI non-blocking for now.

PR #257, opened 2019-09-11
Add opaque key encryption and opaque key decryption

garrickb:
Following the existing `OpaqueSigner` and `OpaqueVerifier`, I have added an `OpaqueKeyEncrypter` and `OpaqueKeyDecrypter`. This adds the ability to provide an opaque encrypter and decrypter. The motivation behind this is to support using a KMS to encrypt/decrypt CEKs.

Related issue: #256

PR #261, opened 2019-09-23
Opaque key encryption/decryption v2 rebase

csstaub:
Via #257


csstaub:
@garrickb We need your signature on the CLA, but otherwise we're good to go.

mbyczkowski:
@csstaub Garrick has signed the CLA already, so we're good here

PR #262, opened 2019-09-23
Move ed25519

csstaub:
Via #258, and also dropping Go <1.13 support on master (no need to support old versions on v3).


mbyczkowski:
@csstaub looks like there's an issue with GolangCI running inside the test suite. Given that we now have GolangCI integration for this repo, maybe we should remove it from the .travis.yml file?

csstaub:
Yeah that makes sense

mbyczkowski:
@csstaub I have rebased this PR on master (and added a little note about Go 1.13+ requirement to README). Does this look good to you?

mbyczkowski:
@csstaub ping on this. Also, should we backport this to v2? I think it could be reasonable to require Go 1.13 at some point for v2, just not sure if we want to do that soon.

csstaub:
@mbyczkowski Let's do it. I can't merge myself due to restricted branch, but feel free to land this. Backporting sounds ok too it's been long enough imo.

mikekamornikov:
@mbyczkowski Do you plan to cut `V2` release with this one anytime soon? We have several deps using `gopkg.in/square/go-jose.v2` and issues like this one blocking `go1.13` upgrade:
```
<our package> imports
  github.com/coreos/go-oidc imports
  gopkg.in/square/go-jose.v2 imports
  golang.org/x/crypto/ed25519 imports
  crypto/ed25519: malformed module path "crypto/ed25519": missing dot in first path element
```

mbyczkowski:
@mikekamornikov I'm not sure this is the same issue. Can you file a new issue with some repro steps?

mikekamornikov:
It looks like our issue was caused by calls from env without `GO111MODULE=on`.

PR #275, opened 2019-11-14
SHA512 is not the same thing as P-521

zamicol:
This is the second time I've seen "521" mistakenly used when referring to SHA-512, so I'm assuming there's a mix up with the P-521 curve and SHA-512

### DIFF
--- a/asymmetric.go
+++ b/asymmetric.go
@@ -507,7 +507,7 @@ func (ctx ecDecrypterSigner) signPayload(payload []byte, alg SignatureAlgorithm)
 		expectedBitSize = 384
 		hash = crypto.SHA384
 	case ES512:
-		expectedBitSize = 521
+		expectedBitSize = 512
 		hash = crypto.SHA512
 	}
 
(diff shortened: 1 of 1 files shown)


zamicol:
I've looked at every other instance of "521" and everyone else seems correct at a glance.

csstaub:
I believe this is actually correct, the ES512 scheme (which is perhaps named confusingly) uses P-521 keys and hashes the data with SHA-512. In this case the expected bit size is to check the key size matches, and P-521 keys are 521 bits long (see line 515).

zamicol:
Ah!  Yes you are totally right.  If you don't mind, I renamed "expectedBitSize" to "expectedCurve" to avoid any future confusion.

asymmetric.go
```diff
@@ -496,24 +496,24 @@ func (ctx edEncrypterVerifier) verifyPayload(payload []byte, signature []byte, a
...
 // Sign the given payload
 func (ctx ecDecrypterSigner) signPayload(payload []byte, alg SignatureAlgorithm) (Signature, error) {
-	var expectedBitSize int
+	var expectedCurve int
 	var hash crypto.Hash
 
 	switch alg {
 	case ES256:
-		expectedBitSize = 256
+		expectedCurve = 256
```
mbyczkowski: might as well add comments for `P-256` and `P-384` for consistency.

mbyczkowski (commented):
It looks like this PR still shows commit 6226d8cb0aa015d8fb6a7efbff25a9fee9ae297d can you squash and rebase or just cherry-pick 0ff51c36eee876ad755c1fac85598c4bccbf460a?

PR #277, opened 2019-11-14
Change phrasing from "full serialization" to "JWE/JWS JSON Serialization"…

zamicol:
… where appropriate as per the RFC's.  

JSON serialization (JWS JSON Serialization and JWE JSON Serialization) is the only option other than compact serialization.  "Full serialization" isn't used in the RFC's and I found the phrasing confusing.  For example, JWT's only use compact serialization, so in this sense compact serialization is "full" since JSON serialization can not be used for JWT's while compact can be used for JWS's, JWE's, and JWT's.  

A great example of the JSON serialization terminology is section 3: https://tools.ietf.org/html/rfc7515#section-3

Here's an excerpt:

> This document defines two serializations for JWSs: a compact, URL-
>    safe serialization called the JWS Compact Serialization and a JSON
>    serialization called the JWS JSON Serialization.  In both
>    serializations, the JWS Protected Header, JWS Payload, and JWS
>    Signature are base64url encoded, since JSON lacks a way to directly
>    represent arbitrary octet sequences.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ The implementation follows the
 [JSON Web Signature](http://dx.doi.org/10.17487/RFC7515) (RFC 7515), and
 [JSON Web Token](http://dx.doi.org/10.17487/RFC7519) (RFC 7519) specifications.
 Tables of supported algorithms are shown below. The library supports both
-the compact and full serialization formats, and has optional support for
+the compact and JWS/JWE JSON Serialization formats, and has optional support for
 multiple recipients. It also comes with a small command-line utility
 ([`jose-util`](https://github.com/square/go-jose/tree/master/jose-util))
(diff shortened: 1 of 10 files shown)


zamicol:
I remembered this morning the biggest reason I found "full" confusing is because compact serialization is "fully" urlbase64 while JSON serialization is only partially urlbase64.  

Using the RFC's terminology clears this confusion.  

Right now this pull request only changes wording in the documentation and comments, but I would recommend changing the function/method names as well.  Would this be a good thing to do for v3 since it would be breaking?

mbyczkowski:
This (the docs/comment change) looks like a reasonable change to me.

Can you open an issue if you want to discuss the API change?

PR #278, opened 2019-11-15
ES256K (secp256k1) signature

### LINKED ISSUE #263: Support JWT with ES256K (secp256k1) signature
As I understand it, go-jose currently supports these keys:

- `P-256`
- `P-384`
- `P-521`

are there any plans to support ES256K keys in go-jose? As in this Java example:
https://connect2id.com/products/nimbus-jose-jwt/examples/jwt-with-es256k-signature

kdimak:
Added ES256K algorithm which uses ECDSA with the secp256k1 curve and the SHA-256 cryptographic hash function.
Fixes #263  in accordance with https://tools.ietf.org/html/draft-ietf-cose-webauthn-algorithms-03 (see [comment](https://github.com/square/go-jose/issues/263#issuecomment-552417473)).


asymmetric_test.go
```diff
@@ -342,6 +342,10 @@ func TestInvalidEllipticCurve(t *testing.T) {
 	if err == nil {
 		t.Error("should not generate ES256 signature with P-384 key")
 	}
+	_, err = signer384.signPayload([]byte{}, ES256K)
```
golangcibot: assignments should only be cuddled with other assignments (from `wsl`)

asymmetric_test.go
```diff
@@ -342,6 +342,10 @@ func TestInvalidEllipticCurve(t *testing.T) {
...
+	if err == nil {
```
golangcibot: only one cuddle assignment allowed before if statement (from `wsl`)

asymmetric_test.go
```diff
@@ -350,6 +354,10 @@ func TestInvalidEllipticCurve(t *testing.T) {
 	if err == nil {
 		t.Error("should not generate ES256 signature with P-521 key")
 	}
+	_, err = signer521.signPayload([]byte{}, ES256K)
```
golangcibot: assignments should only be cuddled with other assignments (from `wsl`)

asymmetric_test.go
```diff
@@ -350,6 +354,10 @@ func TestInvalidEllipticCurve(t *testing.T) {
...
+	if err == nil {
```
golangcibot: only one cuddle assignment allowed before if statement (from `wsl`)

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/square/go-jose/v3
 go 1.12
 
 require (
+	github.com/btcsuite/btcd v0.20.1-beta
```
csstaub: Is there another package implementing `secp256k1` we could use here? I'm a bit wary of having this library depend on the entirety of btcd, it'd be nice if there was just a package for the curve split out somewhere.

csstaub:
Hi @kdimak, thanks for the pull request! I have some concerns about depending on btcd for this, because it's quite a large dependency to force on downstream users. Would prefer if we could use a purpose-built package just for secp256k1.

go.mod
alenhorvat: What about https://github.com/ethereum/go-ethereum/tree/master/crypto/secp256k1? I used this to extend x509 to support der/pem encoding for secp256k1.

csstaub: I think that'd be a better package to use. Requires CGO, so just need to make sure we use build tags to ensure we can still compile with CGO disabled, but I think this is much better than the btcd package.

csstaub: Though also again like the btcd version we'd want to pull in only the secp256k1 implementation as a dep, not the whole of go-ethereum.

alenhorvat: I agree. You'd need only the secp256k1 part, not the whole go-ethereum lib.

alenhorvat: Since this is (at least) a 2nd case where the secp256k1 is needed separately, we could put it in a separate git -- it would require to follow the new updates on go-ethereum.

kdimak:
Hi @csstaub ,  https://github.com/decred/dcrd/tree/master/dcrec was recommended (https://github.com/btcsuite/btcd/issues/1495#issuecomment-571875978) as it's  "maintained by the original btcsuite authors and it provides a module". I've updated the PR to have a dependency on that library solely, please take a look.

cryptosigner/cryptosigner_test.go
```diff
@@ -26,12 +26,13 @@ import (
 	"fmt"
 	"testing"
 
+	btcec "github.com/decred/dcrd/dcrec/secp256k1/v2"
```
csstaub: Rename this import? Since it's not `btcec` anymore, upstream package name is `secp256k1`.

jose-util/generate.go
```diff
@@ -26,19 +26,21 @@ import (
 	"errors"
 	"fmt"
 
+	btcec "github.com/decred/dcrd/dcrec/secp256k1/v2"
```
csstaub: Same comment as above applies here, let's rename the import to `secp256k1`.

jwk.go
```diff
@@ -33,6 +33,7 @@ import (
 	"reflect"
 	"strings"
 
+	btcec "github.com/decred/dcrd/dcrec/secp256k1/v2"
```
csstaub: Again let's rename this import to `secp256k1`.

jwk.go
```diff
@@ -454,6 +455,8 @@ func (key rawJSONWebKey) ecPublicKey() (*ecdsa.PublicKey, error) {
 	switch key.Crv {
 	case "P-256":
 		curve = elliptic.P256()
+	case "secp256k1":
```
csstaub: Huh, is this definitely the right string? Maybe add a comment here and quote the appropriate RFC to make sure people reading this code in the future won't be confused about the dissonance.

shared.go
```diff
@@ -23,6 +23,8 @@ import (
 	"errors"
 	"fmt"
 
+	btcec "github.com/decred/dcrd/dcrec/secp256k1/v2"
```
csstaub: Rename to `secp256k1.

signing_test.go
```diff
@@ -26,6 +26,8 @@ import (
 	"reflect"
 	"testing"
 
+	btcec "github.com/decred/dcrd/dcrec/secp256k1/v2"
```
csstaub: Rename to `secp256k1`.

shared.go
```diff
@@ -495,6 +498,8 @@ func curveName(crv elliptic.Curve) (string, error) {
 		return "P-384", nil
 	case elliptic.P521():
 		return "P-521", nil
+	case btcec.S256():
+		return "secp256k1", nil
```
csstaub: Add comment here, like `// sic!` maybe. Similar to what we do for the 512/521 bits.

jwk.go
kdimak: Yeah, I will add a reference to https://tools.ietf.org/html/draft-ietf-cose-webauthn-algorithms-03

mozillazg:
`LoadPublicKey` and `LoadPrivateKey` don't support PEM/DER-encoded data for ES256K algorithm yet.

https://github.com/square/go-jose/blob/96c717272387d4b4be0da5d3a74a7fb55f3938f7/jose-util/utils.go#L42-L43

https://github.com/square/go-jose/blob/96c717272387d4b4be0da5d3a74a7fb55f3938f7/jose-util/utils.go#L70-L71

go.mod
Baha-sk: @csstaub @alenhorvat https://github.com/decred/dcrd has only crypto stuff, not the whole btcd project, this PR has been updated it.

It's also written 100% in Go, so no C dependencies needed to build it. This is the main reason why we chose this package in this PR. We don't want any external build dependency to ease the building process of go-jose and our projects.

Baha-sk: or a more proper one would be: https://github.com/btcsuite/btcd

marksalpeter:
Two questions:
1. Is there a reason why this isn't merged yet?
2. Is there any possibility of supporting ECDH_ES_A256KW for encryption with secp256k1 ?

PR #279, opened 2019-11-16
[v2] performance: Improve ParseSigned whitespace stripping

pquerna:
- When profiling some other code, I noticed ParseSigned wasn't as fast as I expected.
- Adds a benchmark for ParseSigned
- Replaces regexp to remove whitespace with a string.Builder approach.

`benchcmp` results for `go test -v -run=NONE -bench BenchmarkParseSigned -benchtime=30s -benchmem`
```
benchmark                  old ns/op     new ns/op     delta
BenchmarkParseSigned-4     8184          5461          -33.27%

benchmark                  old allocs     new allocs     delta
BenchmarkParseSigned-4     50             48             -4.00%

benchmark                  old bytes     new bytes     delta
BenchmarkParseSigned-4     3832          3656          -4.59%
```

### DIFF
--- a/encoding.go
+++ b/encoding.go
@@ -23,12 +23,11 @@ import (
 	"encoding/binary"
 	"io"
 	"math/big"
-	"regexp"
+	"strings"
+	"unicode"
 
 	"gopkg.in/square/go-jose.v2/json"
 )
-
-var stripWhitespaceRegex = regexp.MustCompile("\\s")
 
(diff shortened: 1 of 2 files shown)


csstaub:
This is an excellent improvement, thank you @pquerna!

mbyczkowski:
@csstaub any plans to port this to master/v3?

PR #300, opened 2020-04-24
Export jose-util helpers

aeneasr:
The utils for key generations are very helpful in other projects as well. Exporting them seems like a good idea without any downsides as the code is properly tested.

### DIFF
--- a/jose-util/crypto.go
+++ b/jose-util/crypto.go
@@ -16,10 +16,13 @@
 
 package main
 
-import jose "github.com/square/go-jose/v3"
+import (
+	jose "github.com/square/go-jose/v3"
+	"github.com/square/go-jose/jose-util/generator"
+)
 
 func encrypt() {
-	pub, err := LoadPublicKey(keyBytes())
+	pub, err := generator.LoadPublicKey(keyBytes())
 	app.FatalIfError(err, "unable to read public key")
 
 	alg := jose.KeyAlgorithm(*encryptAlgFlag)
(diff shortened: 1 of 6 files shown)


jose-util/go.sum
```diff
@@ -6,6 +6,8 @@ github.com/davecgh/go-spew v1.1.0 h1:ZDRjVQ15GmhC3fiQ8ni8+OwkZQO4DARzQgrnXU1Liz8
 github.com/davecgh/go-spew v1.1.0/go.mod h1:J7Y8YcW2NihsgmVo/mv3lAwl/skON4iLHjSsI+c5H38=
 github.com/pmezard/go-difflib v1.0.0 h1:4DBwDE0NGyQoBHbLQYPwSUPoCMWR5BEzIk/f1lZbAQM=
 github.com/pmezard/go-difflib v1.0.0/go.mod h1:iKH77koFhYxTK1pcRnkKkqfTogsbg7gZNVY4sRDYZ/4=
+github.com/square/go-jose v1.1.2 h1:YCLGvrhulDcHE8AyoqKUKMaY7fQL9tg0c3Uh4QWvQzQ=
```
mbyczkowski, line 9: this seems like an accidental import somewhere?

aeneasr: I'll run go mod tidy again. I didn't change anything in the code base except move the functions around.

aeneasr: Nevermind, you're correct - an incorrect import (without v3) was used somewhere. Nice catch!

aeneasr:
Thank you for the review, I addressed your feedback. Given that we use `cipher`, `cryptosigner`, `jwt` I assumed that `generator.` would be prefered to `generators.`. It's now `generator.NewSigningKey()` - looks pretty sweet to me!

mbyczkowski:
Sorry @aeneasr this got lost in my inbox 😅

aeneasr:
All good, thanks for the merge! :)

Issue #288, opened 2020-02-26
Verification of JWS failed if header contains "crit"


dani2819:
I am generating JSON web signature in JavaScript using node-jws package (https://www.npmjs.com/package/jws). In Headers, I am giving crit: ["exp"] and exp: someTimeStamp. Here is the code used for generating the token:
```
let token = jws.sign({
  header: { alg: 'HS256', crit: ["exp"], exp: Math.floor(Date.now() / 1000) + (60 * 60) },
  payload: "somestring" ,
  privateKey: 'supersecret',
});
```
 In GO, I am verifying it using go-jose package. It verified it correctly, if I don't put crit in headers while creating the token. Otherwise, it won't work saying `square/go-jose: error in cryptographic primitive`. The GO code for verification:
```
import (
    "github.com/square/go-jose"
)

func main() {
    jsonWebSig, err := jose.ParseSigned(token)

    if err != nil {
        panic(err)
    }
    payload, err := jsonWebSig.Verify([]byte("supersecret"))

    fmt.Println(string(payload))
    fmt.Println(err)
}
```

I am unable to figure out the exact problem. Isn't library supports crit in header?

joesiltberg:
I think this is correct behaviour by default. The exp header is not part of standard JWS, and by including it in crit an implementation should fail if it doesn't recognize the header.

I think it should be possible to add extension header attributes though, for instance like 'exp', if you know that you will verify it yourself. go-jose seems to have an internal list of recognized critical headers, but I can't find a way of adding to it.

I have this need as well, and might find the time to put together a pull request to add this functionality, if the maintainers of this project thinks it would be accepted?

PR #301, opened 2020-04-27
Support for deferring critical headers to the caller for validation.

joesiltberg:
Here's an attempt to solve #288.

I've used functional options (https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md) in order to add options to the Verify*-functions without breaking backwards compatibility.

### DIFF
--- a/jwt/builder.go
+++ b/jwt/builder.go
@@ -191,7 +191,7 @@ func (b *signedBuilder) Token() (*JSONWebToken, error) {
 		h[i] = v.Header
 	}
 
-	return b.builder.token(sig.Verify, h)
+	return b.builder.token(sig.Verifier(), h)
 }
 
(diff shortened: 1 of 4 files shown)


joesiltberg:
Note that backwards compatibility is still technically broken since the signature is changed, so this shouldn't be back ported to 2.x

joesiltberg:
@mbyczkowski  @csstaub should I somehow request a review or mark the pull request as ready for review?

Thanks!

csstaub:
Hi @joesiltberg, I'm working on moving dev on the main branch to a new repo in a separate org (will retain v1/2 in this repo, and v3 will be tagged in a new repo that's in a shared org outside of Square). Hang on while I get that set up.

PR #302, opened 2020-04-28
Signing with JWK does not need alg flag

tomoyamachi:
First of all, I provide my deep respect for this project. Great tool.

---

I think we do not need to set `--alg` flag when using JWKs, JWKs already have algorithm property.

### DIFF
--- a/jose-util/crypto.go
+++ b/jose-util/crypto.go
@@ -16,7 +16,11 @@
 
 package main
 
-import jose "github.com/square/go-jose/v3"
+import (
+	"errors"
+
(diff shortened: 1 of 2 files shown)


mbyczkowski:
According to the [RFC 7517](https://tools.ietf.org/html/rfc7517#section-4.4), the `alg` parameter is optional, so you'd probably still have to check that `signingKey.Algorithm` is not empty and require it otherwise.

tomoyamachi:
I understand. Thanks.

Issue #299, opened 2020-04-15
Version 2.5.0: has breaking changes in terms of certificate verfication


hiten1988:
@mbyczkowski 

-Add support for x5u, x5t, and x5t#S256 headers for JWK (#242).

**#issue 1**
has some breaking changes particularly in the jwk.go, where 

**method name** : func (k *JSONWebKey) UnmarshalJSON(data []byte) (err error) 

k.CertificateThumbprintSHA1 = raw.X5tSHA1.bytes()
k.CertificateThumbprintSHA256 = raw.X5tSHA256.bytes()

here the raw value is attempted to convert in bytes but it does not work as expected
in fact, it returns base64 decoded value for kid attribute in the token.

as a result following conditions always fails,

if x5tSHA1Len > 0 && x5tSHA1Len != sha1.Size {
		return errors.New("square/go-jose: invalid JWK, x5t header is of incorrect size")
	}
	if x5tSHA256Len > 0 && x5tSHA256Len != sha256.Size {
		return errors.New("square/go-jose: invalid JWK, x5t header is of incorrect size")
	}

here sha1.Size is the checksum size which is expected to be 20 bytes. but the value which we are assigning k.CertificateThumbprintSHA1 is not a checksum instead it's a raw hash for sha1 or sha256. So when x5tSHA1Len := len(k.CertificateThumbprintSHA1))is executed length turns out to be **40**

solutions:
use x5tSHA1Len := len(sha1.Sum(k.CertificateThumbprintSHA1))

**#issue 2**

// If certificate chain *and* thumbprints are set, verify correctness.
	if len(k.Certificates) > 0 {
		leaf := k.Certificates[0]
		sha1sum := sha1.Sum(leaf.Raw)
		sha256sum := sha256.Sum256(leaf.Raw)
		if len(k.CertificateThumbprintSHA1) > 0 && !bytes.Equal(sha1sum[:], k.CertificateThumbprintSHA1) {
			return errors.New("square/go-jose: invalid JWK, x5c thumbprint does not match x5t value")
		}

		if len(k.CertificateThumbprintSHA256) > 0 && !bytes.Equal(sha256sum[:], k.CertificateThumbprintSHA256) {
			return errors.New("square/go-jose: invalid JWK, x5c thumbprint does not match x5t#S256 value")
		}
	}

I don't understand what are we trying to validate here but this condition seems to be failing always.

feel free to reach out in case of questions or queries

mbyczkowski:
[...]

PR #303, opened 2020-04-28
WIP x5t fix

mbyczkowski:
This is a work-in-progress to fix issue #299

### DIFF
--- a/jwk.go
+++ b/jwk.go
@@ -60,10 +60,10 @@ type rawJSONWebKey struct {
 	Dq *byteBuffer `json:"dq,omitempty"`
 	Qi *byteBuffer `json:"qi,omitempty"`
 	// Certificates
-	X5c       []string    `json:"x5c,omitempty"`
-	X5u       *url.URL    `json:"x5u,omitempty"`
-	X5tSHA1   *byteBuffer `json:"x5t,omitempty"`
-	X5tSHA256 *byteBuffer `json:"x5t#S256,omitempty"`
+	X5c       []string `json:"x5c,omitempty"`
+	X5u       *url.URL `json:"x5u,omitempty"`
+	X5tSHA1   string   `json:"x5t,omitempty"`
+	X5tSHA256 string   `json:"x5t#S256,omitempty"`
 }
 
 // JSONWebKey represents a public or private key in JWK format.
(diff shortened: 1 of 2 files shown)


jwk_test.go
```diff
@@ -354,13 +354,40 @@ func TestInvalidThumbprintsX509(t *testing.T) {
 
 	var jwk2 JSONWebKey
 	err = jwk2.UnmarshalJSON(mismatchedThumbprints)
-	if err == nil {
-		t.Error("should not unmarshal JWK with mismatched thumbprints")
+	if err != nil {
```
mbyczkowski, line 357: this change (and the one below) is only for debugging. Turns out that the tests are passing, because there's another error in the generated JWK (I was expecting only thumbprints to cause errors)

jwk_test.go
```diff
@@ -354,13 +354,40 @@ func TestInvalidThumbprintsX509(t *testing.T) {
...
 
 	err = jwk2.UnmarshalJSON(shortThumbprints)
-	if err == nil {
-		t.Error("should not unmarshal JWK with too short thumbprints")
+	if err != nil {
+		t.Error("should not unmarshal JWK with too short thumbprints", err)
+	}
+}
+
+func TestValidThumbprintsX509(t *testing.T) {
```
mbyczkowski, line 367: this test should really just need a newly generated JWK that's valid (I copied this one from the test above, but the cert from x5c doesn't match the pub key (from RSA params))

jwk.go
```diff
@@ -233,15 +233,29 @@ func (k *JSONWebKey) UnmarshalJSON(data []byte) (err error) {
...
-	k.CertificateThumbprintSHA1 = raw.X5tSHA1.bytes()
-	k.CertificateThumbprintSHA256 = raw.X5tSHA256.bytes()
+
+	// x5t parameters are base64url-encoded SHA thumbprints
+	// See RFC 7517, Section 4.8, https://tools.ietf.org/html/rfc7517#section-4.8
+	x5tSHA1bytes, err := base64.RawURLEncoding.DecodeString(raw.X5tSHA1)
+	if err != nil {
+		return errors.New("square/go-jose: invalid JWK, x5t header has invalid encoding")
+	}
+	k.CertificateThumbprintSHA1 = x5tSHA1bytes
```
mangas, line 252: For me this is still a problem as it doesn't really return 20 bytes but 40, which is always different from sha1.Sum which returns [20]byte. 

The example auth0 keys in the issue also have this problem for me.

mangas: Which also means that line 262 always fails for me

mbyczkowski: ok, so my understanding was that `x5t` is base64url-encoded SHA in raw bytes form, not a hex.

mangas: So for this one I think that was the problem, it's a base64url-encoded SHA hex. Although I don't think this is standard as I understand it, it would still be good to attempt it and fallback to assuming it was raw format if it fails?

mangas: Adding on line 254 works for me, Marshalling still has a problem afterwards where bytes.Equal is returning false for the sha1 check although the values look the same, I'll be able to look into this further tomorrow 
```
	hx, err := hex.DecodeString(string(x5tSHA1bytes))
	if err == nil {
		x5tSHA1bytes = hx
	}
```

mbyczkowski: I'm still pondering this, but if the thumbprint was a hex then it wouldn't need to be base64-encoded.

This might be tangential (since it's not directly discussing `x5t`), but the [JSON Web Key (JWK) Thumbprint](https://tools.ietf.org/html/rfc7638) explicitly mentions base64-encoding raw bytes of the SHA output.

mangas: To be fair the JWK RFC also only mentions Base64 Sha1 for the field specifically
```
base64url-encoded SHA-1 thumbprint
```

The problem right now is that go-auth0 uses go-jose to parse the JWKs and they use hex so although it is a bit strange to include would be great to not break it either if possible

mangas: I guess either that or introducing an option to make this validation optional would work as a non ideal alternative

mbyczkowski:
closing in favor of #304

PR #304, opened 2020-04-29
Fix x5t

mangas:
Fixes https://github.com/square/go-jose/issues/299

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/square/go-jose/v3
 go 1.12
 
 require (
+	github.com/google/go-cmp v0.4.0
 	github.com/stretchr/testify v1.5.1
 	golang.org/x/crypto v0.0.0-20190911031432-227b76d455e7
 )
--- a/jwk.go
+++ b/jwk.go
@@ -27,6 +27,7 @@ import (
 	"crypto/sha256"
 	"crypto/x509"
 	"encoding/base64"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"math/big"
@@ -145,9 +146,12 @@ func (k JSONWebKey) MarshalJSON() ([]byte, error) {
 	if len(k.Certificates) > 0 {
 		expectedSHA1 := sha1.Sum(k.Certificates[0].Raw)
 		expectedSHA256 := sha256.Sum256(k.Certificates[0].Raw)
-		if !bytes.Equal(k.CertificateThumbprintSHA1, expectedSHA1[:]) ||
-			!bytes.Equal(k.CertificateThumbprintSHA256, expectedSHA256[:]) {
-			return nil, errors.New("square/go-jose: invalid SHA-1 or SHA-256 thumbprint, does not match cert chain")
+
+		if len(k.CertificateThumbprintSHA1) > 0 && !bytes.Equal(k.CertificateThumbprintSHA1, expectedSHA1[:]) {
+			return nil, errors.New("square/go-jose: invalid SHA-1 thumbprint, does not match cert chain")
+		}
+		if len(k.CertificateThumbprintSHA256) > 0 && !bytes.Equal(k.CertificateThumbprintSHA256, expectedSHA256[:]) {
+			return nil, errors.New("square/go-jose: invalid or SHA-256 thumbprint, does not match cert chain")
 		}
 	}
 
@@ -240,8 +244,28 @@ func (k *JSONWebKey) UnmarshalJSON(data []byte) (err error) {
 	*k = JSONWebKey{Key: key, KeyID: raw.Kid, Algorithm: raw.Alg, Use: raw.Use, Certificates: certs}
 
 	k.CertificatesURL = raw.X5u
-	k.CertificateThumbprintSHA1 = raw.X5tSHA1.bytes()
-	k.CertificateThumbprintSHA256 = raw.X5tSHA256.bytes()
+
+	// x5t parameters are base64url-encoded SHA thumbprints
+	// See RFC 7517, Section 4.8, https://tools.ietf.org/html/rfc7517#section-4.8
+	x5tSHA1bytes, err := base64.RawURLEncoding.DecodeString(string(raw.X5tSHA1.bytes()))
+	if err != nil {
+		return errors.New("square/go-jose: invalid JWK, x5t header has invalid encoding")
+	}
+
+	fmt.Printf("#### %x\n", x5tSHA1bytes)
+	hx, err := hex.DecodeString(string(x5tSHA1bytes))
+	if err == nil {
+		x5tSHA1bytes = hx
+	}
+	fmt.Printf("#### %x\n", x5tSHA1bytes)
+
+	k.CertificateThumbprintSHA1 = x5tSHA1bytes
+
+	x5tSHA256bytes, err := base64.RawURLEncoding.DecodeString(string(raw.X5tSHA256.bytes()))
+	if err != nil {
+		return errors.New("square/go-jose: invalid JWK, x5t#S256 header has invalid encoding")
+	}
+	k.CertificateThumbprintSHA256 = x5tSHA256bytes
 
 	x5tSHA1Len := len(k.CertificateThumbprintSHA1)
 	x5tSHA256Len := len(k.CertificateThumbprintSHA256)
--- a/jwk.go
+++ b/jwk.go
@@ -61,10 +61,10 @@ type rawJSONWebKey struct {
 	Dq *byteBuffer `json:"dq,omitempty"`
 	Qi *byteBuffer `json:"qi,omitempty"`
 	// Certificates
-	X5c       []string    `json:"x5c,omitempty"`
-	X5u       *url.URL    `json:"x5u,omitempty"`
-	X5tSHA1   *byteBuffer `json:"x5t,omitempty"`
-	X5tSHA256 *byteBuffer `json:"x5t#S256,omitempty"`
+	X5c       []string `json:"x5c,omitempty"`
+	X5u       *url.URL `json:"x5u,omitempty"`
+	X5tSHA1   string   `json:"x5t,omitempty"`
+	X5tSHA256 string   `json:"x5t#S256,omitempty"`
 }
 
 // JSONWebKey represents a public or private key in JWK format.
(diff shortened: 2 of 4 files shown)


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/square/go-jose/v3
 go 1.12
 
 require (
+	github.com/google/go-cmp v0.4.0
```
mangas, line 6: This is more for output readability on the test but happy to drop it if you'd rather not add the dependency

jwk.go
```diff
@@ -129,13 +130,13 @@ func (k JSONWebKey) MarshalJSON() ([]byte, error) {
 		if x5tSHA1Len != sha1.Size {
 			return nil, fmt.Errorf("square/go-jose: invalid SHA-1 thumbprint (must be %d bytes, not %d)", sha1.Size, x5tSHA1Len)
 		}
-		raw.X5tSHA1 = newFixedSizeBuffer(k.CertificateThumbprintSHA1, sha1.Size)
+		raw.X5tSHA1 = base64.RawURLEncoding.EncodeToString(k.CertificateThumbprintSHA1)
```
csstaub, line 133: This could be `newFixedSizeBuffer(k.CertificateThumbprintSHA1, sha1.Size).base64()` which I believe has additional error-checking and verifies the size.

jwk.go
```diff
@@ -129,13 +130,13 @@ func (k JSONWebKey) MarshalJSON() ([]byte, error) {
...
 	}
 	if x5tSHA256Len > 0 {
 		if x5tSHA256Len != sha256.Size {
 			return nil, fmt.Errorf("square/go-jose: invalid SHA-256 thumbprint (must be %d bytes, not %d)", sha256.Size, x5tSHA256Len)
 		}
-		raw.X5tSHA256 = newFixedSizeBuffer(k.CertificateThumbprintSHA256, sha256.Size)
+		raw.X5tSHA256 = base64.RawURLEncoding.EncodeToString(k.CertificateThumbprintSHA256)
```
csstaub, line 139: This could be `newFixedSizeBuffer(k.CertificateThumbprintSHA1, sha1.Size).base64()` which I believe has additional error-checking and verifies the size.

jwk_test.go
```diff
@@ -252,39 +254,114 @@ func TestRoundtripX509(t *testing.T) {
...
+
+			require.Empty(t, cmp.Diff(jsonbar, jsonbar2))
+			if !bytes.Equal(jsonbar, jsonbar2) {
+				t.Error("roundtrip should not lose information")
+			}
+		})
 	}
 }
 
+func TestRoundtripX509Hex(t *testing.T) {
```
csstaub, line 327: Thanks for adding tests!

csstaub:
Looks like there's a build failure but otherwise this seems good.

mangas:
Build failure was the linter rule but IMO reducing complexity will hurt readability in this particular function

jwk.go
mangas: That causes some problems with !bytes.Equal on line 296 that's why I kept this one instead

mangas:
Travis is happy now it seems :)

jwk_test.go
```diff
@@ -252,39 +254,114 @@ func TestRoundtripX509(t *testing.T) {
...
+	],
+	"x5t": "MDYxMjU0ZmRmNzIwZjJjMGU0YmQzZjMzMzlhMmZlNTM1MGExNWRlMQ",
+	"x5t#S256": "MjAzMjRhNGI5MmYxMjI2OGVmOWFlMDI1ZmQ1Yzc5ZDE1OGZmNzQ1NzQwMDkyMTk2ZTgzNTNjMDAzMTUxNzUxMQ"
+}`
+
+	// json output
```
mbyczkowski, line 342: what's the purpose of the json blob below and why are `x5t` headers different than in the input above?

jwk.go
```diff
@@ -240,16 +244,42 @@ func (k *JSONWebKey) UnmarshalJSON(data []byte) (err error) {
...
-	k.CertificateThumbprintSHA256 = raw.X5tSHA256.bytes()
+
+	// x5t parameters are base64url-encoded SHA thumbprints
+	// See RFC 7517, Section 4.8, https://tools.ietf.org/html/rfc7517#section-4.8
+	x5tSHA1bytes, err := base64.RawURLEncoding.DecodeString(raw.X5tSHA1)
+	if err != nil {
+		return errors.New("square/go-jose: invalid JWK, x5t header has invalid encoding")
+	}
+
+	if len(x5tSHA1bytes) > sha1.Size {
```
mbyczkowski, line 255: @mangas we had a bit of a discussion in https://github.com/square/go-jose/pull/303#discussion_r416998839 and I think it would be good to drop a comment here to explain the fallback and how RFC is unclear wether the thumbprints should be raw bytes or a hex string representation. Also if it's a hex then it we know the length -- it would be `2 * sha1.Size` and (and equivalent for SHA256).

mbyczkowski: Also, can we get 2 valid roundtrip tests in which one has `base64(SHA-bytes)` and the other with `base64(SHA-hex)`?

jwk_test.go
mangas: This is basically what the output should look like, I can try to make it comparable maybe by parsing the json I wasn't sure if it was worth doing since this proves the roundtrip

jwk.go
mangas: We have that test, TestRoundtripX509Hex. Takes the hex and produces a non-hex checksum result.

mangas: Cool, I'm make the size checks more strict and add a comment.

jwk_test.go
mangas: I've fixed this up to use the expected output. The output was kept assuming that regardless of the input encoding, the output is always a base64 encoded version of byte format checksum

jwk.go
mangas: Let me know if this comments is clear enough

go.mod
mbyczkowski: seems fine to me

jwk.go
```diff
@@ -240,16 +244,46 @@ func (k *JSONWebKey) UnmarshalJSON(data []byte) (err error) {
...
+	if err != nil {
+		return errors.New("square/go-jose: invalid JWK, x5t header has invalid encoding")
+	}
+
+	// RFC 7517, Section 4.8 is ambiguous as to whether the digest output should be byte or hex,
+	// for this reason, after base64 decoding, if the size is sha1.Size it's likely that the value is a byte encoded
+	// checksum so we skip this. Otherwise if the checksum was hex encoded we expect a 40 byte sized array so we'll
+	// try to hex decode it. When Marshalling this value we'll always use a base64 encoded version of byte format checksum.
+	if len(x5tSHA1bytes) == 2*sha1.Size {
+		if hx, err := hex.DecodeString(string(x5tSHA1bytes)); err == nil {
```
mbyczkowski, line 260: if we can't decode hex, we should error here (especially that we know the length is not a correct SHA1 length) and same for SHA256 hex

mbyczkowski: oh I see, we compare the lengths below and then fail. Seems ok to do it this way. Maybe a small comment could help, though

go.mod
mbyczkowski: especially that Go modules knows when an import is used in tests, and even though this dep is present in go.mod, it will not bloat the actual code.

jwk_test.go
```diff
@@ -252,37 +254,117 @@ func TestRoundtripX509(t *testing.T) {
...
+			jwk: JSONWebKey{
+				Key:                       testCertificates[0].PublicKey,
+				KeyID:                     "bar",
+				Algorithm:                 "foo",
+				Certificates:              testCertificates,
+				CertificateThumbprintSHA1: x5tSHA1[:],
+			},
+		},
+		{
+			name: "no x5t25",
```
mbyczkowski, line 292: small nit: this should probably say `no x5t`?

jwk_test.go
```diff
@@ -252,37 +254,117 @@ func TestRoundtripX509(t *testing.T) {
...
+			name: "no optional x5ts",
+			jwk: JSONWebKey{
+				Key:          testCertificates[0].PublicKey,
+				KeyID:        "bar",
+				Algorithm:    "foo",
+				Certificates: testCertificates,
+			},
+		},
+		{
+			name: "no x5t",
```
mbyczkowski, line 282: small nit: this should probably say `no x5t#S256`?

jwk.go
mangas: Fixed with an explicit message here

jwk_test.go
mangas: fixed

jwk_test.go
mangas: fixed

jwk.go
```diff
@@ -240,16 +244,51 @@ func (k *JSONWebKey) UnmarshalJSON(data []byte) (err error) {
...
+	}
+
+	// RFC 7517, Section 4.8 is ambiguous as to whether the digest output should be byte or hex,
+	// for this reason, after base64 decoding, if the size is sha1.Size it's likely that the value is a byte encoded
+	// checksum so we skip this. Otherwise if the checksum was hex encoded we expect a 40 byte sized array so we'll
+	// try to hex decode it. When Marshalling this value we'll always use a base64 encoded version of byte format checksum.
+	if len(x5tSHA1bytes) == 2*sha1.Size {
+		hx, err := hex.DecodeString(string(x5tSHA1bytes))
+		if err != nil {
+			return fmt.Errorf("unable to hex decode x5t, %w", err)
```
mbyczkowski, line 262:
```suggestion
			return fmt.Errorf("square/go-jose: invalid JWK, unable to hex decode x5t: %w", err)
```

jwk.go
```diff
@@ -240,16 +244,51 @@ func (k *JSONWebKey) UnmarshalJSON(data []byte) (err error) {
...
+
+	x5tSHA256bytes, err := base64.RawURLEncoding.DecodeString(raw.X5tSHA256)
+	if err != nil {
+		return errors.New("square/go-jose: invalid JWK, x5t#S256 header has invalid encoding")
+	}
+
+	if len(x5tSHA256bytes) == 2*sha256.Size {
+		hx256, err := hex.DecodeString(string(x5tSHA256bytes))
+		if err != nil {
+			return fmt.Errorf("unable to hex decode x5t#S256, %w", err)
```
mbyczkowski, line 278:
```suggestion
			return fmt.Errorf("square/go-jose: invalid JWK, unable to hex decode x5t#S256: %w", err)
```

PR #315, opened 2020-06-17
jwt: Claims.Validate: Clarify that you must set e.Time

evanj:
The default arguments to Validate ignore the timestamps in the JWT.
This makes sense, but is a dangerous default, since users may not
realize that the timestamps in their JWTs are not being checked.
Clarify the documentation to try and make this error less common.

### DIFF
--- a/jwt/validation.go
+++ b/jwt/validation.go
@@ -47,6 +47,8 @@ func (e Expected) WithTime(t time.Time) Expected {
 
 // Validate checks claims in a token against expected values.
 // A default leeway value of one minute is used to compare time values.
+// Set e.Time to the current time, otherwise the Expiry, IssuedAt and
+// NotBefore claims will be ignored.
 //
 // The default leeway will cause the token to be deemed valid until one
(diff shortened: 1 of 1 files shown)


evanj:
I actually would suggest changing the default behavior: If you call Validate, it *should* validate the timestamps by default. This is pretty important to getting JWT security right, since otherwise a leaked token is valid forever. I also literally just made this mistake. Another alternative could be to make it an error. I'm happy to submit a PR to do either of these. Since v3 is still not "released" I think a breaking change like that would be okay?

In the rare case a user really does want to ignore the timestamps, which should be very rare, they can always set `Claims.Expiry` and other fields to `nil` before calling `Validate`.

shaxbee:
Could we default to time.Now() if time is zero? This would provide safe fallback.

csstaub:
I'd be ok with changing the default behavior here.

evanj:
I will work on a PR for that then.

evanj:
PR #317 changes the default. I think it should be merged instead of this one, so I'm going to close this.

PR #317, opened 2020-06-18
jwt: Claims.Validate: Treat zero Expected.Time as now

evanj:
This ensures we always validate the exp/iat/nbf fields. Previously,
if the caller forgot to set the Time field, they were ignored. This is
a very easy and critical security bug. Users who want to ignore these
fields can still set the corresponding Claims fields to nil.

### DIFF
--- a/jwt/validation.go
+++ b/jwt/validation.go
@@ -25,7 +25,9 @@ const (
 )
 
 // Expected defines values used for protected claims validation.
-// If field has zero value then validation is skipped.
+// If field has zero value then validation is skipped, with the exception of
+// Time, where the zero value means "now." To skip validating them, set the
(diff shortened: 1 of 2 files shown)


evanj:
I closed #315, since it just clarified the documentation. I think this is a better fix, but is a breaking change. If we don't want to change the API, we should merge the documentation fix instead.

csstaub:
Tagging @mbyczkowski for merge and cherry-pick on v2 (since I don't have write access to the master branch in this repo anymore)

mbyczkowski:
@csstaub do you think we should backport this to v2 given that this is a breaking change?

csstaub:
Merged this into the v3 branch in go-jose/go-jose. Thanks @evanj!

PR #338, opened 2020-11-06
fix: add ExtraHeaders from options object to multi encrypter

### LINKED ISSUE #337: Using MultiEncrypter with ContentType
Right now in NewMultiEncrypter the ExtraHeaders from the options object is never copied to the encrypter. This means you can't make an encrypter that will satisfy the NestedBuilder requirement of having a "JWT" content type.

Here is an example of v2 and this behavior:
https://play.golang.org/p/2iQoF2myz9k

Here is an example of v3 showing this behavior:
https://play.golang.org/p/kAb6M2m0_Gr

Is there something I'm missing or is this expected behavior for some reason?

kelsin:
Attempts to fix #337 in a very naive way.

### DIFF
--- a/crypter.go
+++ b/crypter.go
@@ -216,6 +216,7 @@ func NewMultiEncrypter(enc ContentEncryption, rcpts []Recipient, opts *Encrypter
 
 	if opts != nil {
 		encrypter.compressionAlg = opts.Compression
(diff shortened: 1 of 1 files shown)


csstaub (approved):
Thanks @kelsin!

PR #341, opened 2020-11-19
Add support for non-pointer jose.JSONWebKeySet

leobrines:
I found that it throws an error when you don't send it as a pointer and later I saw that it was not supported in the code, which seemed strange to me because to sign [JSONWebKey pointer or non-pointer](https://github.com/square/go-jose/blob/5c87222f69b2401c305bcb103b875524e13de347/signing.go#L173) are supported. So I just add it :)


csstaub (approved):

csstaub:
Thanks @leobrines!

Issue #351, opened 2021-05-24
json.Unmarshal converts json literal integer to float64 instead of int64 when taget type is of type interface{}


narg95:
# Issue

`json.Unmarshal`  unmarshals integers values into `float64` when target type is of type `interface{}`.

# Expected behavior

If a json literal is assignable to an integer without losing precision the  `json.Unmarshal` must convert to `int64` instead of `float64`

The problem is located in `json/decode.go` in [convertNumber function](https://github.com/square/go-jose/blob/v2/json/decode.go#L732):

```golang
func (d *decodeState) convertNumber(s string) (interface{}, error) {
	if d.useNumber {
		return Number(s), nil
	}
	f, err := strconv.ParseFloat(s, 64)  // See here! By default it converts to Float
	if err != nil {
		return nil, &UnmarshalTypeError{"number " + s, reflect.TypeOf(0.0), int64(d.off)}
	}
	return f, nil
}
```

by default it does `strconv.ParseFloat`

# Steps to Reproduce

```go
func TestUnmarshalNumber(t *testing.T) {
	var v interface{}
	exp := int64(1621867181)
	err := json.Unmarshal([]byte(`1621867181`), &v)
	require.NoError(t, err)
	require.IsType(t, exp, v)
}
```

Error:
```bash
=== RUN   TestUnmarshalNumber
    token_test.go:502:
        	Error Trace:	token_test.go:502
        	Error:      	Object expected to be of type int64, but was float64
        	Test:       	TestUnmarshalNumber
--- FAIL: TestUnmarshalNumber (0.00s)
FAIL
FAIL	token/jwt	0.025s
```

# Proposed Solution

`convertNumber` should try first `strconv.ParseInt` and in case of `err` it does `strconv.ParseFloat`

If you agree with the proposal I can submit a PR

csstaub:
Sounds reasonable. What does the latest encoding/json in Go's standard library do in this case?

nestorvw:
Also `float64` https://play.golang.org/p/8JKQ3G8nr1M but would'nt it make sense that the custom `go-jose`'s json marshalling library addresses `jwt` specific needs like a correct data type for claims with timestamps e.g. `iat, exp, nbf`? otherwise it could just use standard golang library.

narg95:
[...]

PR #352, opened 2021-05-25
Unmarshal JSON integer numbers into `interface{}`  as `int64`

narg95:
# Related Issue

#351

# Description

- allow to unmarshal JSON integer numbers into an `interface{}` as an `int64`
- add `Decoder.SetNumberType` method to configure the desired number unmarshal behavior: `float64`, `json.Number`, `int64 / float64`
- deprecate `Decoder.UseNumber`


json/decode.go
```diff
@@ -723,16 +736,24 @@ func (d *decodeState) literal(v reflect.Value) {
...
+
 	f, err := strconv.ParseFloat(s, 64)
 	if err != nil {
 		return nil, &UnmarshalTypeError{"number " + s, reflect.TypeOf(0.0), int64(d.off)}
 	}
+
+	// if it has no decimal value use int64
+	if d.numberType == UnmarshalIntOrFloat {
+		if fi, fd := math.Modf(f); fd == 0.0 {
+			return int64(fi), nil
```
csstaub, line 754: If we use fi here rather than re-parsing from the string, won't we possibly lose precision? We'd cap out at 9,007,199,254,740,992 vs. the 18,446,744,073,709,551,615 that a uint64 could represent.

narg95: good catch, the maximum safe integer in float representation is `2^53` but the maximum `int64` is 2^63-1
btw. you mentioned `uint64` but this PR is not meant for supporting unsigned integers but `int64`.

narg95: fixed

csstaub (approved):
Looks good. Thanks for the contribution!

narg95:
you're welcome, thanks to your for reviewing it

narg95:
do you think we can release it? `2.5.2` or `2.6.0` ?

csstaub:
I'll release this in the v3 branch in github.com/go-jose/go-jose

PR #358, opened 2021-07-24
RFC: Support parsing without payload

asac:
Hi Square,

According to JWS spec one can produce payload-detached serializations in compact _and_ JSON format, but jose currently does not allow me to parse a JSON document that does not have payload included.... In our scenario we will need the protected info before validation and having this would be convenient  ....

 - asac

p.s.  Note: I realize taht tests are failing, but wanted to discuss approach first ....


csstaub:
Hi @asac, can you re-open this diff on the go-jose/go-jose repo for the v3 branch? This repo (square/go-jose) is frozen at the v2 branch and will only accept security fixes for v2, no new features.

asac:
@csstaub Yes, I would, but I cannot find the v3 branch in this repo :/.

asac:
ok, i think i found it in new repo ...https://github.com/go-jose/go-jose/tree/v3

PR #369, opened 2022-03-26
Fix Allow multiple audiences

### LINKED ISSUE #286: Multiple audiences and JWT validation
Looking at the code for audience validation:

	if len(e.Audience) != 0 {
		for _, v := range e.Audience {
			if !c.Audience.Contains(v) {
				return ErrInvalidAudience
			}
		}
	}

It looks like the intention there is to require the JWT to have *all* of the expected audiences, and fails if one is missing from the JWT, is that correct?

For example, if the JWT has an audience claim of `https://my-resource-server`, and the resource server has an audience configuration of `[ "https://my-resource-server", "https://my-other-server"]` then the validation will fail, because the JWT does not include `https://my-other-server` as an audience claim.

Other implementations I have used approach multiple audiences on the resource server side as a "can be any one of these" validation check, rather than "must be all of these" validation check, which would allow the resource server to serve content for multiple audiences.

Is the current approach intentional or would you be open to a PR changing its behaviour?

shnmorimoto:
## Related Issue

fixed #286 

## Description
- allow any match for multiple audiences
- currently go-jose requires all match for multiple audiences

### DIFF
--- a/jwt/validation.go
+++ b/jwt/validation.go
@@ -87,10 +87,15 @@ func (c Claims) ValidateWithLeeway(e Expected, leeway time.Duration) error {
 	}
 
 	if len(e.Audience) != 0 {
+		flag := false
 		for _, v := range e.Audience {
-			if !c.Audience.Contains(v) {
-				return ErrInvalidAudience
+			if c.Audience.Contains(v) {
+				flag = true
+				break
 			}
+		}
+		if !flag {
+			return ErrInvalidAudience
 		}
 	}
 
--- a/jwt/validation_test.go
+++ b/jwt/validation_test.go
@@ -44,6 +44,17 @@ func TestFieldsMatch(t *testing.T) {
 		assert.NoError(t, c.Validate(v))
 	}
 
+	claimsWithSingleAudience := Claims{
+		Issuer:   "issuer",
+		Subject:  "subject",
+		Audience: []string{"a1"},
+		ID:       "42",
+	}
+
+	for _, v := range valid {
+		assert.NoError(t, claimsWithSingleAudience.Validate(v))
+	}
+
 	invalid := []struct {
 		Expected Expected
 		Error    error


shnmorimoto:
@csstaub Could you please review this PR? If the mention is wrong, I'm sorry.

I know that v3 repository already exists.
However, I think this is a bug fix, so I raised the PR to this repository.

csstaub:
I can review this, but I don't have permissions on this repository anymore @shnmorimoto so it's not possible for me to do a release with this.

csstaub:
If you raise this on the v3 repo I can merge it there and release it as a minor version

shnmorimoto:
@csstaub Thanks for replying!
I understood the current situation.

I raised PR to v3 repository.
https://github.com/go-jose/go-jose/pull/10

shnmorimoto:
@mbyczkowski If the mention is wrong, I apologize.

This fix already had been merged on v3 repository.
But, it seems v2 version is still being used by many users and libraries.
So, if possible, could you merge this fix and release it on this repository?

shnmorimoto:
@mbyczkowski Sorry for bothering you.
Could you confirm this PR?

RJNY:
Bump - I can confirm that we're also running into this issue on our team using v2. Would love to see this merged

Jthomas54:
Ditto, it is stopping the use of this package.

csstaub:
Is this still an issue with v3: https://github.com/go-jose/go-jose ?

csstaub:
The square/go-jose repo is the v2 version, v3 development moved over to go-jose/go-jose

Jthomas54:
Looks like v3 will solve it but noticed that it is a dependency of  `github.com/auth0/go-jwt-middleware/v2/jwks` so doubtful they will move to the "unstable" v3.

csstaub:
There's a stable release for v3 now so I'd recommend to switch and submit new changes there. I don't have access to this repo anymore since I left Square. I'll see if we can update the README here.

